PR #31, opened 2019-07-18
adding docs to master

kenotron:


kenotron:
gonna handle this with #32 instead

PR #104, opened 2019-08-12
Improve package list formatting

ecraig12345:
When listing packages, put each one on a new line so it's easier to read.

Also add a VS Code settings.json with basic formatting options.


ecraig12345:
@kenotron Please look at this? :)

packages/beachball/src/getChangedPackages.ts
```diff
@@ -72,11 +72,8 @@ export function getChangedPackages(branch: string, cwd: string, fetch: boolean)
   });
 
   if (changeFilePackageSet.size > 0) {
-    console.log(
-      `Your local repository already has change files for these packages: ${[...changeFilePackageSet]
-        .sort()
-        .join(', ')}`
-    );
+    console.log('Your local repository already has change files for these packages:' +
```
kenotron: This will insert the comma character before each line. Let ’s  join it with new line character instead maybe?

kenotron:
thx for merging!

PR #116, opened 2019-08-21
Integration tests

jdhuntington:
This tests only the most basic scenario, but is a starting place.

Next steps:
- validate more complicated scenarios
- clean up temp folders after test runs

### DIFF
--- a/packages/beachball/src/fixtures/repository.ts
+++ b/packages/beachball/src/fixtures/repository.ts
@@ -1,0 +1,144 @@
+import { exec as nativeExec } from "child_process";
+import * as process from "process";
+import * as tmp from "tmp";
+import { writeFile } from "fs";
+import { promisify } from "util";
+
+const writeFileAsync = promisify(writeFile);
+
+const packageJson = JSON.stringify({
+  name: "foo"
+});
+
+async function dirAsync(options: tmp.DirOptions): Promise<tmp.DirResult> {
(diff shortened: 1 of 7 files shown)


packages/beachball/src/fixtures/repository.ts
```diff
@@ -0,0 +1,118 @@
...
+export class RepositoryFactory {
+  root?: tmp.DirResult;
+
+  async create(): Promise<void> {
+    this.root = await dirAsync({ prefix: "beachball-repository-upstream-" });
+    await runInDirectory(this.root!.name, [
+      "git init",
+      "touch README",
+      "git add README",
+      'echo \'{ "name": "foo" }\' > package.json',
```
kenotron: does this work across all platforms?

jdhuntington: I am guessing this will not work on windows.

packages/beachball/src/fixtures/repository.ts
```diff
@@ -112,7 +124,7 @@ export class Repository {
     await runInDirectory(this.root.name, [
       `touch ${newFilename}`,
       `git add ${newFilename}`,
-      `git commit -m '${newFilename}'`
+      `git commit --author=\"CI User <user@example.com>\" -m '${newFilename}'`
```
kenotron: do you still need \" inside a ` ?

Issue #115, opened 2019-08-21
Publish command should verify the private field before attempting to publish


kenotron:
The status of a package.json could actually have changed since the change file has been created, so we need to verify this before publishing. We should still do the bumping - just maybe skip the publish of the package for simplicity.

PR #117, opened 2019-08-21
Don't publish private packages

leeviana:
Right now we assume that all valid packages mentioned in change files should be published to, but some of the packages - though valid - are private and should not be valid. This check filters our operational package information to only public ones.


leeviana:
Fixes #115

Issue #128, opened 2019-08-28
Scenario testing for publishing


kenotron:
@jdhuntington added the scenario testing infra for us doing change check, creating change files, etc. There are some fixtures about publish scenario tests that we need:

1. local npm registry via https://verdaccio.org/. 
2. BARE git repo (git init --bare) so that git push would work

We should assert aspects about a package after package has been published to a verdaccio via `npm pack`

PR #136, opened 2019-09-04
Add publish tests using verdaccio.

JasonGore:
Partially addresses #128.

Add publish tests by using verdaccio to simulate a registry for publishing and testing against. Verdaccio is also configured to run in memory, allowing for simple and effect state reset between tests with no persistence.

The really difficult part of implementing this PR was a solution that would scale with tests running in parallel using a test registry. Each parallel test needs its own instance of a test registry with a known state and its own port.

The PR does not have multiple tests making use of the test registry, but this scaling capability can be seen simply by copying and pasting `packageManager.test.ts` a few times to see that they all run correctly without colliding or failing.


packages/beachball/package.json
```diff
@@ -17,6 +17,7 @@
     "test-watch": "jest --watch"
   },
   "dependencies": {
+    "find-free-port": "~2.0.0",
```
kenotron: probably just need to be a devDep in this case.

PR #139, opened 2019-09-10
Fixes paths in beachball to use full paths in the package info

kenotron:
This aids testing so that we do not use hard coded paths to represent package.json locations.


package.json
```diff
@@ -8,10 +8,10 @@
   "license": "MIT",
   "scripts": {
     "build:all": "lerna run build --stream",
-    "build": "lerna run build --scope beachball",
-    "test": "lerna run test --scope beachball",
+    "build": "yarn workspace beachball build",
+    "test": "yarn workspace beachball test",
     "postbuild:all": "rimraf docs && ncp packages/website/public docs",
-    "start": "lerna run start",
+    "start": "yarn workspace beachball start",
```
JasonGore: Curious what drove this just to know if I should make this change in flamegrill too.

kenotron: This is a realization that most of the time i'm only building ONE package. I also don't expect us to have more than one package in this repo... so Lerna isn't really needed honestly.

PR #159, opened 2019-09-25
Add option to specify a defaultNpmTag on a per package basis

acoates-ms:


packages/beachball/src/cli.ts
```diff
@@ -46,7 +46,7 @@ const options: CliOptions = {
   publish: args.publish === false ? false : true,
   push: args.push === false ? false : true,
   registry: args.registry || 'https://registry.npmjs.org/',
-  tag: args.tag || 'latest',
```
kenotron: If we don't default to add a latest, npm would publish "latest" fine, but then this clause in the "createTag" will end up stamping "latest" to git as well:

```ts
  if (tag !== 'latest') {
    createTag(tag, cwd);
  }
```

Maybe an easy fix is to have `defaultNpmTag` to default to `latest`

kenotron:
Thanks for adding those tests

PR #168, opened 2019-10-04
Bump specific change

kenotron:
In this change, when we merge, bump, push; we would reuse the existing bumpInfo from when the code was checked out during a build release job. This way, new change files would not be deleted. They would be preserved, making the merge more resilient to race conditions.

Also, when a change file is created (manually) for a private package, this will just only warn instead of blowing up the publish.

Lastly, I've added a check to make sure that if someone were to add change files that reference incorrect package names that it'll not bomb out but just ignores those files.


packages/beachball/src/__tests__/publishGit.test.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+      repo2.rootPath
+    );
+
+    git(['push', 'origin', 'master'], { cwd: repo2.rootPath });
+
+    // 4. Pretend to continue on with repo1's publish
+    await bumpAndPush(bumpInfo, publishBranch, options);
+
+    // 5. In a brand new cloned repo, make assertions
```
JasonGore: 👍

PR #169, opened 2019-10-04
Ignore private package publishing

kenotron:
Through some manual testing, private package publishes causes the publish to try to delete the change files for those private packages still. We should skip those actions as well as tagging.


JasonGore:
Hmm, are there any tests covering these changes?

kenotron:
I don't properly have an e2e test covering private package publish issues.

PR #200, opened 2019-11-13
Add --bump-deps flag to bump all dependent packages

### LINKED ISSUE #189: Packages whose dependencies change should be published
Currently If I have 2 packages:

Package A: 
```
{
  dependencies: {
    packageB: "1.0.3"
  }
}
```

And I publish package B, it correctly updates the dependencies of package A, but it doesn't publish package A with the new dependencies.  This means users of package A will never see the new version of package B.

We should have some kind of option to configure `beachball` to be more aggressive about publishing updates when their dependencies change.

statm:
Given 3 packages: A <- B <- C, when package A revs, current bump behavior will result in:
```
A: bump,
B: dependency entry A updated,
C: unchanged.
```

With `--bump-deps` flag, beachball will bump all of A's direct and indirect dependents:
```
A: bump,
B: bump (patch), dependency entry A updated,
C: bump (patch), dependency entry B updated.
```

Note: as @kenotron points out in the comments below, currently all dependent packages' version will be bumped as a 'patch' (0.0.**1**). This does align with rush's approach, but is not necessarily the best strategy.

All bumped packages will also be published during `beachball publish`.

Resolves #189


packages/beachball/src/bump.ts
```diff
@@ -79,11 +80,48 @@ export function performBump(
...
+        }
+
+        const info = packageInfos[pkgName];
+        const packageJsonPath = info.packageJsonPath;
+        const packageJson = JSON.parse(fs.readFileSync(packageJsonPath).toString());
+
+        const allDeps = { ...packageJson.dependencies, ...packageJson.devDependencies };
+        for (const dep of Object.keys(allDeps)) {
+          if (bumpedPackages.includes(dep)) {
+            packageJson.version = semver.inc(packageJson.version, "patch");
```
kenotron: Therein lies the question: do we bump a patch for the indirect bumps? What if the bumped version was a minor, do we bump patch in this? Also, what if in a single patch bump, the dep's external dep got bumped a major version - what do we do then?

There's a lot of strategies here - this is a good start though. I HOPE this is good enough for 80% case.

statm: Good catch.

I've thought about this, and I think the best approach will be to **not** use `--bump-deps` flag, but run `beachball change; beachball bump` multiple times and manually choose each package's bump type. We'll test this flag in one or two of our internal repos and see how things go.

Also updated PR desc mentioning this point.

PR #209, opened 2019-11-26
Bump peerDependencies

statm:
If a package is referenced in another's `peerDependencies`, it should also be bumped.


packages/beachball/src/__tests__/publishGit.test.ts
```diff
@@ -11,6 +11,10 @@ import { ChangeInfo } from '../ChangeInfo';
 describe('publish command', () => {
   let repositoryFactory: RepositoryFactory;
 
+  beforeAll(() => {
```
kenotron: Is this necessary?

PR #266, opened 2020-03-03
This adds a --scope for check and change that can be applied to repos that manages several groups of packages

kenotron:
In this change, we added a param called "scope". For now, it applies to change and check so that those commands can make checks / change files only for scoped packages.

For example: 

```
beachball change --scope packages/grouped/*
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
         "--runInBand",
         "--config",
         "${workspaceRoot}/packages/beachball/jest.config.js",
-        "${fileBasename}"
+        "${fileBasenameNoExtension}}"
       ],
       "windows": {
         "args": [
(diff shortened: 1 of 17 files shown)


.vscode/launch.json
```diff
@@ -24,7 +24,33 @@
...
+          "${fileBasenameNoExtension}"
+        ]
+      },
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen"
+    },
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug Jest E2E Tests",
```
xugao: Let's update the task above with `"name": "Debug Jest Unit Tests"`

packages/beachball/src/monorepo/getScopedPackages.ts
```diff
@@ -0,0 +1,22 @@
...
+import minimatch from 'minimatch';
+import path from 'path';
+
+export function getScopedPackages(options: BeachballOptions) {
+  const packageInfos = getPackageInfos(options.path);
+  if (!options.scope) {
+    return Object.keys(packageInfos);
+  }
+
+  const scopes = options.scope!.map(scope => path.join(options.path, scope));
```
xugao: `!` should not be needed here?

packages/beachball/src/__e2e__/monorepo/getScopedPackages.test.ts
```diff
@@ -0,0 +1,22 @@
...
+describe('getScopedPackages', () => {
+  it('can scope packages', async () => {
+    const repoFactory = new MonoRepoFactory();
+    await repoFactory.create();
+    const repo = await repoFactory.cloneRepository();
+
+    const scopedPackages = getScopedPackages({
+      path: repo.rootPath,
+      scope: ['packages/grouped/*'],
+    } as BeachballOptions);
```
xugao: we should avoid casting here and other places too. we can change to `getScopedPackages(options: Partial<BeachballOptions>`

.vscode/launch.json
kenotron: yes!

PR #279, opened 2020-03-20
Fix promptChangeFiles to honor disallowedChangeTypes defined in version groups

xugao:
Currently when prompting change type for change files, we only filter using `disallowedChangeTypes` defined at top level. Instead we should be looking at the `disallowedChangeTypes` defined in version groups as well.


packages/beachball/src/changefile/promptForChange.ts
```diff
@@ -91,3 +94,18 @@ export async function promptForChange(options: BeachballOptions) {
 
   return packageChangeInfo;
 }
+
+function getDisallowedChangeTypes(
```
kenotron: It seems like this is a good start - where the group disallowedChangeTypes will override the setting for the individual packages.

PR #280, opened 2020-03-20
Fix version group bumping with none change type

xugao:
when a package inside a group is bumped with `none` change type, it's always bumped as `patch` due to we use `patch` as default in `updateRelatedChangeType`

also fixed default `dependentChangeType` value in change file. if change is `none`, do not set `dependentChangeType` to `patch`

### DIFF
--- a/packages/beachball/src/__tests__/bump/updateRelatedChangeType.test.ts
+++ b/packages/beachball/src/__tests__/bump/updateRelatedChangeType.test.ts
@@ -78,7 +78,7 @@ describe('updateRelatedChangeType', () => {
     expect(bumpInfo.packageChangeTypes['bar']).toBe('minor');
   });
 
-  it('should bump all packages in a group together', () => {
+  it('should bump all packages in a group together as minor', () => {
     const bumpInfo = _.merge(_.cloneDeep(bumpInfoFixture), {
       dependentChangeTypes: {
         foo: 'minor',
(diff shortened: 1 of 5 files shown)


packages/beachball/src/__tests__/bump/updateRelatedChangeType.test.ts
```diff
@@ -102,6 +102,81 @@ describe('updateRelatedChangeType', () => {
...
+    expect(bumpInfo.packageChangeTypes['unrelated']).toBeUndefined();
+  });
+
+  it('should bump all packages in a group together as none with dependents', () => {
+    const bumpInfo = _.merge(_.cloneDeep(bumpInfoFixture), {
+      dependents: {
+        foo: ['bar'],
+      },
+      dependentChangeTypes: {
+        foo: 'none',
```
xugao: if pkgs inside a group take dependency on each other and this is `patch` (which is what we set as default when all change files), we still get `patch` bump for all packages inside the group despite the only change inside the group is `none` type

packages/beachball/src/bump/updateRelatedChangeType.ts
```diff
@@ -22,15 +22,15 @@ export function updateRelatedChangeType(
 
   const disallowedChangeTypes = packageInfos[pkgName].options?.disallowedChangeTypes ?? [];
 
-  let depChangeType = getMaxChangeType('patch', dependentChangeTypes[pkgName], disallowedChangeTypes);
+  let depChangeType = getMaxChangeType(MinChangeType, dependentChangeTypes[pkgName], disallowedChangeTypes);
```
kenotron: The "dependencyChangeType" in change file generation should be "none" when the change type is "none"

PR #282, opened 2020-03-23
Infer commit hash from change file introduction commit

ecraig12345:
Currently beachball records the most recent commit hash when a change file is created, but that causes problems in some cases, such as if you rebase after creating the change file. It also makes it harder to link a change file to a particular GitHub PR since the referenced commit may not exist in history if the repo squashes PRs.

This PR removes the commit hash from the change files, and instead infers it while generating the changelog, based on the actual commit hash that introduced the change file.

Also add a call to `tmp.setGracefulCleanup()` to ensure test repos are actually cleaned up.

### DIFF
--- a/packages/beachball/src/fixtures/repository.ts
+++ b/packages/beachball/src/fixtures/repository.ts
@@ -76,10 +76,7 @@ export class Repository {
     this.origin = path;
   }
 
-  /**
-   * @returns The hash of the new commit
-   */
-  async commitChange(newFilename: string, content?: string): Promise<string> {
+  async commitChange(newFilename: string, content?: string): Promise<void> {
     if (!this.root) {
       throw new Error('Must initialize before cloning');
     }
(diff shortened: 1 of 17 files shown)


packages/beachball/src/git/index.ts
```diff
@@ -202,6 +198,22 @@ export function getCurrentHash(cwd: string) {
   return null;
 }
 
+/**
+ * Get the commit hash in which the file was first added.
+ */
+export function getFileAddedHash(filename: string, cwd: string) {
```
xugao: we can call existing function `getCurrentHash` for getting the hash

packages/beachball/src/types/ChangeInfo.ts
```diff
@@ -5,9 +5,12 @@ export interface ChangeInfo {
   comment: string;
   packageName: string;
   email: string;
-  commit: string;
   date: Date;
   dependentChangeType?: ChangeType;
 }
 
-export type ChangeSet = Map<string, ChangeInfo>;
+export interface FullChangeInfo extends ChangeInfo {
```
xugao: how about rename `ChangeInfo` to `ChangeFileInfo` and keep `ChangeInfo` to be the one contains `commit`

packages/beachball/src/git/index.ts
ecraig12345: No, that's for getting the most recent commit hash, which might not be the one where the changefile was added.

ecraig12345: This is being run when the changelog is generated, not when the change file is generated

packages/beachball/src/fixtures/repository.ts
```diff
@@ -76,7 +76,10 @@ export class Repository {
     this.origin = path;
   }
 
-  async commitChange(newFilename: string, content?: string) {
+  /**
+   * @returns The hash of the new commit
+   */
+  async commitChange(newFilename: string, content?: string): Promise<string> {
```
xugao: revert this? i don't think you are using the return value

PR #288, opened 2020-03-25
Changelog: improve package changelog by leveraging github and tag

xugao:
### Change descriptions
1. Added `github` to beachball options to collect basic GitHub info (e.g repo, owner)
2. Made version header a link using tag
3. Added "Compare changes" link using tag

**Before change:**
## 7.105.3
Wed, 25 Mar 2020 23:28:35 GMT

**After change:**
## [7.105.3](https://github.com/microsoft/fluentui/tree/office-ui-fabric-react_v7.105.3)
Wed, 25 Mar 2020 23:28:35 GMT
[Compare changes](https://github.com/microsoft/fluentui/compare/office-ui-fabric-react_v7.105.2..office-ui-fabric-react_v7.105.3)

### DIFF
--- a/packages/beachball/src/changelog/renderPackageChangelog.ts
+++ b/packages/beachball/src/changelog/renderPackageChangelog.ts
@@ -1,10 +1,30 @@
-import { PackageChangelog, ChangelogEntry } from '../types/ChangeLog';
+import { PackageChangelog, ChangelogEntry, ChangelogJsonEntry } from '../types/ChangeLog';
 import _ from 'lodash';
+import { GitHubInfo } from '../types/BeachballOptions';
 
-export function renderPackageChangelog(changelog: PackageChangelog, isGroupedChangelog: boolean = false) {
+export interface PackageChangelogRenderOptions {
+  changelog: PackageChangelog;
+  isGroupedChangelog: boolean;
+  previousChangelogEntry: ChangelogJsonEntry | undefined;
+  github?: GitHubInfo;
+}
+
+export function renderPackageChangelog(options: PackageChangelogRenderOptions): string {
+  const { changelog, isGroupedChangelog, previousChangelogEntry, github } = options;
+  const header =
+    github && changelog.tag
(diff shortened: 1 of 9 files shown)


packages/beachball/src/types/BeachballOptions.ts
```diff
@@ -39,10 +39,21 @@ export interface RepoOptions {
   changehint: string;
   disallowedChangeTypes: ChangeType[] | null;
   defaultNpmTag: string;
+  github?: GitHubInfo;
```
kenotron: Let's not make this Github specific?

packages/beachball/src/changelog/renderPackageChangelog.ts
```diff
@@ -1,10 +1,30 @@
...
+  const { changelog, isGroupedChangelog, previousChangelogEntry, github } = options;
+  const header =
+    github && changelog.tag
+      ? `[${changelog.version}](https://github.com/${github.owner}/${github.repo}/tree/${changelog.tag})`
+      : changelog.version;
+
+  let subHeader = changelog.date.toUTCString();
+  subHeader +=
+    github && previousChangelogEntry?.tag
+      ? `\n[Compare changes](https://github.com/${github.owner}/${github.repo}/compare/${previousChangelogEntry?.tag}..${changelog.tag})`
```
kenotron: We should definitely not make this so github specific. This is where I wished we had a custom renderer function that folks can use. One way to work around having a custom function would be to allow for a template string or a function...

ecraig12345: If we do have a custom render function, maybe we could provide an OOB one (or an option) which works with github

ecraig12345: This will also be a design issue relevant to the feature I'm working on, retrieving and adding PR links to changelogs.

xugao: @kenotron @ecraig12345 
I went with supporting custom render function: https://github.com/microsoft/beachball/pull/289/files

let me know what you think

PR #289, opened 2020-03-26
Support custom render for package changelog

xugao:
Support custom render for markdown content of each new package version inside CHANGELOG.md

See how this will be used by fluent ui: 
https://github.com/microsoft/fluentui/pull/12443

### DIFF
--- a/packages/beachball/src/changelog/writeChangelog.ts
+++ b/packages/beachball/src/changelog/writeChangelog.ts
@@ -8,8 +8,9 @@ import { renderChangelog } from './renderChangelog';
 import { renderJsonChangelog } from './renderJsonChangelog';
 import { BeachballOptions } from '../types/BeachballOptions';
 import { isPathIncluded } from '../monorepo/utils';
-import { PackageChangelog } from '../types/ChangeLog';
+import { PackageChangelog, ChangelogJsonEntry, ChangelogJson } from '../types/ChangeLog';
 import { mergeChangelogs } from './mergeChangelogs';
+import { renderPackageChangelog } from './renderPackageChangelog';
 
 export function writeChangelog(
   options: BeachballOptions,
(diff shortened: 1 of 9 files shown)


packages/beachball/src/changelog/writeChangelog.ts
```diff
@@ -105,17 +124,31 @@ function writeChangelogFiles(changelog: PackageChangelog, changelogPath: string,
...
-      ? JSON.parse(fs.readFileSync(changelogJsonFile).toString())
-      : { entries: [] };
-    const nextJson = renderJsonChangelog(previousJson, changelog);
-    fs.writeFileSync(changelogJsonFile, JSON.stringify(nextJson, null, 2));
-  } catch (e) {
-    console.warn('The CHANGELOG.json file is invalid, skipping writing to it', e);
+    const renderCustomPackageChangelog = options.changelog?.renderPackageChangelog;
+    let renderPackageChangelog;
+    if (renderCustomPackageChangelog) {
+      renderPackageChangelog = () => {
```
ecraig12345: It would be nice if we could figure out an API for letting people do *some* modifications to the changelog without having to reinvent all the logic. Not sure exactly what that would look like though. Some ways it could be broken down:
- custom entry renderer
- custom section header renderer
- plus a full custom renderer option

xugao: if need, i would give back default renderers for different sections: header/entry header/entry body. but they are not helpful for our use case right now. so i'd like to avoid until there is a need

ecraig12345: I think actually it would be helpful now with your PR to Fabric--that way we don't have to entirely reinvent the wheel when all we want to customize is the section header so far.

xugao: originally i was thinking giving `renderChangelogEntries` as default renderer, but the moment you add PR link, you can't use it any more

ecraig12345: Right, that's why I think we should have multiple custom renderers. It's a little more complex but not terrible.

PR #297, opened 2020-03-28
Add custom render functions for changelog parts

ecraig12345:
Add fine-grained custom render functions for the changelog markdown. This way if someone wants to just modify one little part, they don't have to re-create the entire rendering logic. 

Variant of @xugao's PR #289.

Example:
```js
module.exports = {
  // ...
  changelogOptions: {
    customRenderers: {
      renderHeader: (renderInfo) => {
        return [
          `## ${renderInfo.newPackageVersion.version}`, 
          renderInfo.newPackageVersion.date.toUTCString(),
          `[Compare changes](${/*get compare link*/})`
        ].join('\n');
      },
      renderEntry: (entry, renderInfo) =>
        `- ${entry.comment} (${entry.author}, PR #${/*get PR number*/})`
    }
  }
};
```

Also add the git tag to CHANGELOG.json, which was part of Xu's change.

### DIFF
--- a/packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
+++ b/packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
@@ -1,4 +1,5 @@
-import { MarkdownChangelogRenderOptions, renderChangelog } from '../../changelog/renderChangelog';
+import * as _ from 'lodash';
+import { MarkdownChangelogRenderOptions, renderChangelog, markerComment } from '../../changelog/renderChangelog';
 
 const previousContent = `# Change Log - foo
 
@@ -7,6 +8,22 @@ This log was last generated on Wed, 21 Aug 2019 21:20:40 GMT and should not be m
 ## 1.2.0
 
 (content here)
+`;
+
+const previousContentWithMarker = `# Change Log - foo
+
+This log was last generated on Wed, 21 Aug 2019 21:20:40 GMT and should not be manually modified.
+
+${markerComment}
+
+## 1.2.0
+
+(content here)
+`;
+
+const badPreviousContent = `# Change Log - foo
+
+This log was last generated on Mon, 02 Mar 2020 12:25:44 GMT and should not be manually modified.
 `;
 
 describe('renderChangelog', () => {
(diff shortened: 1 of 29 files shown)


ecraig12345:
@xugao and @kenotron this is ready for review! I changed the API a little and added tests (mostly snapshots because that lets you see the actual output and ensure it looks good).

packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      previousContent,
+      previousJson: {} as any,
+      changelogOptions: {},
+    };
+  }
+
+  it('handles no previous content', () => {
+    const options = getOptions();
+    options.previousContent = '';
+    expect(renderChangelog(options)).toMatchSnapshot();
```
xugao: thanks for adding comprehensive tests! (great idea using snapshot)

we have some tests in changelog.test.ts too in writeChangelog, could you change to snapshot as well?

packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
```diff
@@ -0,0 +1,85 @@
...
+            renderInfo.newVersionChangelog.date.toUTCString(),
+            `[Compare changes](http://real-github-compare-link)`,
+          ].join('\n');
+        },
+        renderEntry: (entry, renderInfo) => `- ${entry.comment} (${entry.author}, PR #123)`,
+      },
+    };
+
+    const result = renderChangelog(options);
+    expect(result).toContain('Compare changes');
```
kenotron, line 68: doesn't the match snapshot already perform "toContain" checks more or less?

packages/beachball/src/changelog/renderChangelog.ts
```diff
@@ -1,12 +1,56 @@
...
+export function renderChangelog(renderOptions: MarkdownChangelogRenderOptions): string {
+  const {
+    previousJson,
+    previousContent,
+    newVersionChangelog,
+    isGrouped,
+    changelogOptions: { renderPackageChangelog: customRenderPackageChangelog, customRenderers },
+  } = renderOptions;
+
+  const previousLogEntries = previousContent ? previousContent.substring(previousContent.indexOf('##')) : '';
```
kenotron, line 19: This seems arbitrary now with a custom renderer - not sure you can count on that? This needs to be fixed before we can merge this I believe.

kenotron: At the moment I think maybe the best is to create a marker with a markdown comment that isn't rendered:
https://stackoverflow.com/questions/4823468/comments-in-markdown

packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
ecraig12345: I just wanted to add a couple explicit checks as well in case snapshot updates get ignored. Hopefully updates will be harder to ignore in this case than the giant DOM snapshots in Fabric, but better to be safe?

packages/beachball/src/changelog/renderChangelog.ts
ecraig12345: Do you specifically want one of those special faux-comment types or is a standard HTML `<!-- comment -->` okay? (I think an HTML comment is fine.)

ecraig12345: On a related note...oops... https://github.com/microsoft/fluentui/blob/master/packages/keyboard-key/CHANGELOG.md#change-log---fluentuikeyboard-key-1

packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
ecraig12345: I updated them to only use snapshots. If you want me to also keep any of the explicit tests just let me know and I'll add it back.

packages/beachball/src/changelog/writeChangelog.ts
```diff
@@ -8,37 +8,39 @@ import { renderChangelog } from './renderChangelog';
 import { renderJsonChangelog } from './renderJsonChangelog';
 import { BeachballOptions } from '../types/BeachballOptions';
 import { isPathIncluded } from '../monorepo/utils';
-import { PackageChangelog } from '../types/ChangeLog';
+import { PackageChangelog, ChangelogJson } from '../types/ChangeLog';
 import { mergeChangelogs } from './mergeChangelogs';
 
-export function writeChangelog(
+export async function writeChangelog(
```
xugao, line 14: we also call this in `performBump`. need to make sure we change that chain of callers to async too

packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
xugao: looks great. thank you!

packages/beachball/src/changelog/writeChangelog.ts
ecraig12345: fixed, thanks!

ecraig12345:
I added a marker comment `<!-- Start content -->` which marks the start of the latest generated content. Also realized I needed to make all the renderers async in case they need to do network calls (like fetching stuff from github).

packages/beachball/src/__tests__/changelog/renderChangelog.test.ts
kenotron: sure.

Issue #275, opened 2020-03-19
ChangeLog: Support tagging in change file


xugao:
Currently in change file, we collect:
1. Change type: patch, minor, major, none
2. comment: author to describe the change

We should support a way for author to future group change into a particular genre (tag), they can be:
- Breaking change
- Bug fixe
- Feature
- Documentation
- Performance

PR #298, opened 2020-03-30
Change file prompt: support customizing prompt questions

xugao:
Enable customizing the prompt with the following:
1. description for each change type 
2. add change area prompt

For instance:
Beachball options:
```

  changeFilePrompt: {
    changePrompt: prompt => {
      const { changeType } = prompt;
      changeType.choices &&
        changeType.choices.forEach(choice => {
          if (choice.value === 'patch') {
            choice.title = ' [1mPatch[22m      - Changes that bumps packge with patch version.';
          } else if (choice.value === 'minor') {
            choice.title = ' [1mMinor[22m      - Changes that bumps packge with minor version.';
          }
        });

      const changeAreaPrompt = {
        type: 'select',
        name: 'area',
        message: 'Change area',
        choices: [
          { value: 'fix', title: 'Bug fix' },
          { value: 'perf', title: 'Performance' },
          { value: 'doc', title: 'Documentation' },
        ],
      };
      return [prompt.changeType, changeAreaPrompt, prompt.description];
    },
  },
```
Will lead to prompt:
image
image

No change to existing description prompt

Addresses #275

### DIFF
--- a/change/beachball-2020-03-30-13-44-57-xgao-custom-change-type-prompt.json
+++ b/change/beachball-2020-03-30-13-44-57-xgao-custom-change-type-prompt.json
@@ -1,0 +1,8 @@
+{
+  "type": "minor",
+  "comment": "Change file prompt: support customizing prompt questions",
+  "packageName": "beachball",
+  "email": "xgao@microsoft.com",
+  "dependentChangeType": "patch",
+  "date": "2020-03-30T20:44:57.439Z"
+}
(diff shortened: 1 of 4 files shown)


kenotron:
I was thinking that we would go even farther and letting the end user pass in a function to take over the entire process of prompt. So you would have users pass you a:

```js
// beachball.config.js

{
   changePrompt: (prompt) => {
       // do what you will with prompt
   }
}
```

Internally, it would make sure that the end result / response contain certain valid fields like changeType, description, packageName, etc.

PR #299, opened 2020-03-30
Adds an option to allow package publishes to be retried

### LINKED ISSUE #290: Retry publishing if one package fails
Right now if publishing one package in a monorepo fails, beachball doesn't retry it, which results in a really messy state that has to be fixed manually. Instead it should automatically retry at least once (could add a retry count option).

kenotron:
This fixes #290 - sometimes publishes fail due to network conditions. This retry will add resiliency to the publish process.

### DIFF
--- a/packages/beachball/src/__e2e__/publishRegistry.test.ts
+++ b/packages/beachball/src/__e2e__/publishRegistry.test.ts
@@ -21,6 +21,57 @@ describe('publish command (registry)', () => {
     await registry.reset();
   });
 
+  it('will perform retries', async () => {
+    registry.stop();
+
+    const repositoryFactory = new RepositoryFactory();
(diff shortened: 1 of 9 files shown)


packages/beachball/src/__e2e__/publishRegistry.test.ts
```diff
@@ -21,6 +21,57 @@ describe('publish command (registry)', () => {
     await registry.reset();
   });
 
+  it('will perform retries', async () => {
```
xugao: add check for retry is actually happening?

kenotron: sure

ecraig12345:
Thanks for doing this!

Issue #29, opened 2019-07-17
Feature: Handle partial publishes


kenotron:
Handle partial publishes:

minispec:

- add a "sync" command that can synchronize a partially updated git repo with the current versions of the published npm packages. 
- ~delete change files for a subset of these~
- ~should mark partially upgraded versions to be unpublished and set latest dist-tag to the last known good~
- ~on publish AGAIN, make sure to skip over unpublished versions and apply a patch (with a changelog entry) about the skipped version~

kenotron:
@leeviana @christiango - please do chime in to see if you can give more detail to the spec.

leeviana:
Discussed a similar problem today and came up with another potential solution: retry fetch/push step (since this is what often fails) when there are issues.

kenotron:
@dzearing - this came up again when npm had an outage during our publish time.

kenotron:
The mini spec is too magical. We cannot guarantee these actions will address the issues. What we should do is only do the first line (sync command)

PR #300, opened 2020-03-30
adding a sync command to help recover

kenotron:
Fixes a long standing feature request: #29 

In the event of a partial publish failure, it is tedious to have to find out which package versions to sync. This automates the procedure. This script does NOT unpublish old partially published versions, however.

### DIFF
--- a/packages/beachball/src/help.ts
+++ b/packages/beachball/src/help.ts
@@ -21,13 +21,14 @@ Commands:
   changelog           - based on change files, create changelogs and then unlinks the change files
   bump                - bumps versions as well as generating changelogs
   publish             - bumps, publishes to npm registry (optionally does dist-tags), and pushes changelogs back into master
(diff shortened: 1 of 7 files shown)


packages/beachball/src/help.ts
```diff
@@ -21,13 +21,14 @@ Commands:
   changelog           - based on change files, create changelogs and then unlinks the change files
   bump                - bumps versions as well as generating changelogs
   publish             - bumps, publishes to npm registry (optionally does dist-tags), and pushes changelogs back into master
+  sync                - synchronizes published versions of packages from a registry, makes local package.json changes to match what is published
```
xugao: should also add to:
https://github.com/microsoft/beachball/blob/master/README.md
https://github.com/microsoft/beachball/blob/master/packages/beachball/README.md

kenotron: good point! added!

PR #313, opened 2020-04-10
Formatting and typing cleanup

ecraig12345:
Some typing and formatting fixes I noticed would be helpful while doing other work on beachball:
- Enable `noImplicitAny` and `noUnusedLocals` and make related typing fixes
- Update prettier to handle new TS syntax
- Fix prettier commands and run them
- Manual formatting fixes: add blank lines; remove empty param tags
- Use `fs-extra` (and its promise-based async methods) everywhere

### DIFF
--- a/packages/beachball/src/bump/setDependentsInBumpInfo.ts
+++ b/packages/beachball/src/bump/setDependentsInBumpInfo.ts
@@ -1,15 +1,15 @@
 import { BumpInfo } from '../types/BumpInfo';
+import { PackageDeps } from '../types/PackageInfo';
+
 /**
  * Gets dependents for all packages
  *
  * Example: "BigApp" deps on "SomeUtil", "BigApp" would be the dependent
- *
- * @param bumpInfo
  */
 export function setDependentsInBumpInfo(bumpInfo: BumpInfo): void {
   const { packageInfos, scopedPackages } = bumpInfo;
   const packages = Object.keys(packageInfos);
-  const dependents = {};
+  const dependents: BumpInfo['dependents'] = {};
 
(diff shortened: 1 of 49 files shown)


packages/beachball/src/bump/setDependentsInBumpInfo.ts
```diff
@@ -19,8 +19,9 @@ export function setDependentsInBumpInfo(bumpInfo: BumpInfo): void {
     const info = packageInfos[pkgName];
     const depTypes = ['dependencies', 'devDependencies', 'peerDependencies'];
     depTypes.forEach(depType => {
-      if (info[depType]) {
-        for (let [dep, _] of Object.entries(info[depType])) {
+      const deps: PackageDeps | undefined = (info as any)[depType];
+      if (deps) {
+        for (let dep of Object.keys(deps)) {
```
xugao, line 24: directly use `for .. in`: `for (let depPackage in deps)`

ecraig12345: I thought that was still considered bad practice due to `for..in` including non-own properties

xugao: hm i never heard of it as bad practice. in this use case, i dont think there is a real concern?

ecraig12345: We have a lint rule against it in Fabric (unless accompanied by a hasOwnProperty check), and my understanding is that it's a common lint rule in other projects too. Object.keys is much more commonly used than for..in across both beachball and oufr.

PR #323, opened 2020-04-17
Publish packages in the right order based on dependency graph

### LINKED ISSUE #322: Publish dependencies first then dependents
If package A depends on package B, we should publish B first then A.

xugao:
Fixes #322

## Problem
if dependent publishes first and dependencies failed to publish, we ended up publishing packages which are not installable for users.

## Solution
Do a topological sort on the packages before publishing them to guarantee dependencies are published before dependent.

### DIFF
--- a/packages/beachball/src/publish/toposortPackages.ts
+++ b/packages/beachball/src/publish/toposortPackages.ts
@@ -1,10 +1,12 @@
-import { PackageInfos } from '../types/PackageInfo';
-import toposort from 'toposort';
 import * as _ from 'lodash';
+import toposort from 'toposort';
+import { PackageInfos } from '../types/PackageInfo';
 
 /**
  * Topological sort the packages based on its dependency graph.
  * Dependency comes first before dependent.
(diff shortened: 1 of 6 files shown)


packages/beachball/src/publish/toposortPackages.ts
```diff
@@ -0,0 +1,46 @@
...
+ * Dependency comes first before dependent.
+ * @param packages Packages to be sorted.
+ * @param packageInfos PackagesInfos for the sorted packages.
+ */
+export function toposortPackages(packages: string[], packageInfos: PackageInfos): string[] {
+  const packageSet = new Set(packages);
+  const dependencyGraph: [string | undefined, string][] = [];
+
+  packages.forEach(pkgName => {
+    let allDeps: string[] = [];
```
kenotron, line 16: Possibly important in bigger repos - but you can use a Set to get uniques without having to deal with using lodash for it... Sets are also iterable so you shouldn't have to convert it back to an array

PR #324, opened 2020-04-17
validate private package not being a dependency before publish

### LINKED ISSUE #321: Validate package dependencies not being private packages
before we publish, we should validate if package contains deps which are private. to prevent mistakes like: https://github.com/microsoft/fluentui/commit/e7d03cc930fa094538adf9d272b312e21ed934b5

xugao:
Fixes #321

### DIFF
--- a/packages/beachball/src/publish/validatePackageDependencies.ts
+++ b/packages/beachball/src/publish/validatePackageDependencies.ts
@@ -1,4 +1,3 @@
-import * as _ from 'lodash';
 import { BumpInfo } from '../types/BumpInfo';
 import { shouldPublishPackage } from './shouldPublishPackage';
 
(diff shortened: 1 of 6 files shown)


packages/beachball/src/publish/validatePackageDependencies.ts
```diff
@@ -0,0 +1,46 @@
...
+    ['dependencies', 'devDependencies'].forEach(depKind => {
+      const deps = packageInfo[depKind];
+      if (deps) {
+        const depPkgNames = Object.keys(deps);
+        allDeps = allDeps.concat(depPkgNames);
+      }
+    });
+  });
+
+  allDeps = _.uniq(allDeps);
```
kenotron, line 33: Same as the other PR - you can use Set here instead for uniq...

PR #326, opened 2020-04-17
Change prepublish hook to be file based and happen per-package

JasonVMo:
## Removal of old prepublish hook

The first iteration on the prepublish hook allowed for modifying bumpInfo.  This happened during publish before packages were bumped.  While it allowed for modifying versions the implementation comes with some issues:

- it only worked for certain fields like version, changes to other information wouldn't be reflected in the files
- it happens before the last batch of changes are committed to the package.json files, this means that modification in bumpInfo, were they to be respected, might conflict
- it leaks too much in the way of internal implementation details
- it doesn't address the desire to make modifications to package.json that only apply to published files

Eventually the old prepublish hook could reappear as a prebump hook with a more deterministic API such as one that has a specific set of actions which can be taken on the packages.

## New prepublish hook

The new hook happens:

- after bumps have been done
- after updated package.json files have been committed to git
- after we have determined which packages should be published
- before each actual publish command is executed

This allows for modifications to the files before they are published to the registry.  This could be used for:

- modifying package.json files
- cleaning up internal only files
- copying data from elsewhere that should be published together but doesn't need to be in the internal environment.

This also removes the capability for returning a promise.  Because the publish routine is happening immediately after all async operations related to this call should be resolved before it returns.

### DIFF
--- a/packages/beachball/src/publish/publishToRegistry.ts
+++ b/packages/beachball/src/publish/publishToRegistry.ts
@@ -35,7 +35,7 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
 
   const packagesToPublish = toposortPackages([...modifiedPackages, ...newPackages], packageInfos);
 
-  packagesToPublish.forEach(pkg => {
+  for (const pkg of packagesToPublish) {
     const { publish, reasonToSkip } = shouldPublishPackage(bumpInfo, pkg);
     if (!publish) {
       console.log(`Skipping publish - ${reasonToSkip}`);
@@ -47,7 +47,10 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
 
     // run the prepublish hook once, after version bumping but before actually executing npm publish (with retries)
     if (prepublishHook) {
-      prepublishHook(path.dirname(packageInfo.packageJsonPath), packageInfo.name, packageInfo.version);
+      const maybeAwait = prepublishHook(path.dirname(packageInfo.packageJsonPath), packageInfo.name, packageInfo.version);
+      if (maybeAwait instanceof Promise) {
+        await maybeAwait;
+      }
     }
 
     let result;
@@ -72,5 +75,5 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
     displayManualRecovery(bumpInfo, succeededPackages);
     console.error(result.stderr);
     throw new Error('Error publishing, refer to the previous error messages for recovery instructions');
-  });
+  }
 }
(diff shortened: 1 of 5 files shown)


packages/beachball/src/publish/publishToRegistry.ts
```diff
@@ -13,15 +14,7 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
...
-
-  // Execute prepublish hook if available
-  if (options.hooks?.prepublish) {
-    const maybePromise = options.hooks.prepublish(bumpInfo);
-
-    if (maybePromise instanceof Promise) {
-      await maybePromise;
-    }
-  }
+  const prepublishHook = options.hooks?.prepublish;
```
kenotron, line 17: can we preserve the ability to do things with a promise? this way, we give a bit more flexibility to what can be done inside the hook

packages/beachball/src/types/BeachballOptions.ts
```diff
@@ -53,12 +52,13 @@ export interface RepoOptions {
...
-     * the changes will be reverted before pushing
+     * Runs for each package after version bumps have been processed and committed to git, but before the actual
+     * publish command.
      *
-     * This hook expects manipulation to the bumpInfo object (side effects)
+     * This allows for file modifications which will be reflected in the published package but not be reflected in the
+     * repository.
      */
-    prepublish?: (bumpInfo: BumpInfo) => void | Promise<void>;
+    prepublish?: (packagePath: string, name: string, version: string) => void;
```
kenotron, line 61: this should return optionally a promise

packages/beachball/src/publish/publishToRegistry.ts
```diff
@@ -42,7 +35,7 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
 
   const packagesToPublish = toposortPackages([...modifiedPackages, ...newPackages], packageInfos);
 
-  packagesToPublish.forEach(pkg => {
+  for (const pkg of packagesToPublish) {
```
kenotron, line 38: let's do all the prepublish hooks for all packages in one loop, and then loop thru the packages again to publish. If we do this in phases, then we can catch prepub hook errors and prevent any publishes. partial publishes is SO BAD.

packages/beachball/src/publish/publishToRegistry.ts
JasonVMo: Updated

packages/beachball/src/publish/publishToRegistry.ts
JasonVMo: Great idea!  Updated.

PR #328, opened 2020-04-20
Fix multi-package publishing regression (#327).

### LINKED ISSUE #327: Beachball stops publishing to the registry after the first package successfully published.
In this section: https://github.com/microsoft/beachball/blob/9f79bfd9c7b1f656a9277237129faf8297f43084/packages/beachball/src/publish/publishToRegistry.ts#L68-L72

The `return` at line 71 will terminate the entire `publishToRegistry` procedure after the first package is successfully published, leaving a messy state silently if you have multiple packages need to publish.

reli-msft:
Fixes #327 .

The current `return` when after package is successfully published will stop Beachball to publish other packages if multiple packages needs to publish to the registry. This change fixes this regression introduced by #299 .

### DIFF
--- a/change/beachball-2020-04-20-01-01-13-reli-fix-multiple-package-publishing.json
+++ b/change/beachball-2020-04-20-01-01-13-reli-fix-multiple-package-publishing.json
@@ -1,0 +1,8 @@
+{
+  "type": "patch",
+  "comment": "Fix multi-package publishing regression (#327).",
+  "packageName": "beachball",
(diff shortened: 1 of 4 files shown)


kenotron:
Actually the bug was introduced here: https://github.com/microsoft/beachball/commit/9f79bfd9c7b1f656a9277237129faf8297f43084

Where the loop was changed from forEach to a for() loop so the return was returning the wrong closure...

PR #332, opened 2020-04-22
Sync should limit to current major version and skip prerelease

ecraig12345:
When syncing to the latest published version, beachball should only sync to latest of the **same major version** specified in package.json. (Example: Say Fabric 6 publish fails partway through. If I run `beachball sync` before this change, all the packages will be updated to version 7!)

It's probably also best not to try to sync packages which are on a prerelease version, since semver isn't so well-defined in that case and we can't automatically figure out what's next.

### DIFF
--- a/packages/beachball/src/commands/sync.ts
+++ b/packages/beachball/src/commands/sync.ts
@@ -14,17 +14,28 @@ export async function sync(options: BeachballOptions) {
       continue;
     }
 
-    const npmArgs = ['view', pkg, 'version'];
+    // Get the latest of *this major version* of the package.
+    // To do this we have to get all versions within this major version then take the latest.
+    const npmArgs = ['view', `${pkg}@${semver.major(info.version)}`, 'version', '--json'];
     if (options.registry) {
       npmArgs.push('--registry');
       npmArgs.push(options.registry);
     }
     const result = npm(npmArgs);
-    const publishedVersion = result.stdout;
+    let publishedVersion = '';
+
+    try {
+      // The returned array of versions will already be sorted and not include prerelease because
+      // prerelease versions don't satisfy semver pkg@x.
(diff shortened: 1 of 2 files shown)


packages/beachball/src/commands/sync.ts
```diff
@@ -14,17 +14,33 @@ export async function sync(options: BeachballOptions) {
       continue;
     }
 
-    const npmArgs = ['view', pkg, 'version'];
+    if (semver.prerelease(info.version)) {
+      console.log(`Warning: cannot sync prerelease package "${pkg}@${info.version}"`);
```
xugao, line 18: can you elaborate why we cannot sync prerelease?

ecraig12345: Because as far as I'm aware you can't tell in an automated way which prerelease version comes after the current one.

ecraig12345: Also the right thing to do in that situation is unclear: do they want the next prerelease version (which again we can't determine), or the next non-prerelease?

xugao: i'm confused. 
1. are you saying we cannot compare which version is newer on this line:
```
semver.lt(info.version, publishedVersion)
```
2. sync is to sync to the published version. so whatever version it's currently published in npm, we update local version to that if local version is older

kenotron: Thinking about this some more and came to this conclusion: the sync is doing too much here. When I designed this command, the idea is to have it sync down whatever is the @latest dist-tag from npm registry regardless of whatever format it is. This command is really meant as a way to reset your repo to whatever state is captured inside a npm registry. We should actually REMOVE logic regarding `semver.lt()` below.

ecraig12345:
I'm going to just close this for now and leave it for reference in case someone wants to come back later.

PR #346, opened 2020-06-02
Add autoTag option to control git auto created tags (ie: foo_v1.0.0)

ahmedkotb:
- Added a new configuration value to control the creation of git auto tags, (eg: foo_v2.1.0)
the default value for autoTag is true, so no change in configuration is required for existing repos.
- Added unit tests for `tagPackages.ts`
- Updated docs
- Added change file


packages/beachball/src/__e2e__/publishE2E.test.ts
```diff
@@ -62,6 +62,7 @@ describe('publish command (e2e)', () => {
       bumpDeps: true,
       push: true,
       registry: registry.getUrl(),
+      autoTag: true,
```
kenotron, line 65: Per an offline discussion: I think we should have this be called "gitTag" and by default it should be `undefined` which would mean that it would automatically be using "tag". In the future, we'll fill this API out with a function `(pkg, version) => string` that will generate customizable git tags.

ahmedkotb: renamed to `gitTags`  with cli aliases `--git-tags` and `--no-git-tags`

PR #356, opened 2020-06-23
Allow Individual Packages to Opt Out of Git Tags

NickGerleman:
This builds upon #346 and exposes gitTags as an option in package.json. This is useful for monorepos which only want one package to show up in their release view.

This is populated according to the existing rules of `getPackageOptions`, meaning the option we will see precedence of:
1. Cli options
2. Local package
3. Root package
4. Default options (tag)


NickGerleman:
@kenotron Did some end to end testing. Confirmed --no-git-tags still works, we tag by default, and we can specialize behavior per-package. I did need to make a fix for the scneario of a root package disabling git tags, but an individual package enabling, where before we never evaluated logic because we were only checking root/cli settings.

I can't merge this myself, but I think we should be all good.

PR #367, opened 2020-07-20
Sync deps to parallelize (for speed), uses execa (for parallelization), and updates dependent ranges

### LINKED ISSUE #347: Beachball sync should update package references within the same repo
Beachball sync has been a super useful tool for our team when the publish loop fails partway through. One improvement that could be made to make this flow even better would be to update the package references within the same repo.

Say we have package A and package B. Package B depends on the latest of package A in the repo. Package B has an entry `A: ^1.0.0` in package.json. Package A version 1.0.1 was published, but the rest of the publish loop failed, preventing package A's version entry to be updated to 1.0.1. We run beachball sync in this case which update Package A's version to 1.0.1, but does note update Package B's dependency from `A: ^1.0.0` to `A: ^1.0.1`. Adding this capability will make fixing publish failures even easier!

kenotron:
Fixes #347 

This is a refactor of beachball bump & sync commands so that the underlying bits are a bit more aligned with each other. This allows us to have dependent version range bumps to occur during `beachball sync`. Furthermore, we switched to using `execa` to allow npm version commands to occur in parallel (they're not like publish that is order dependent).


christiango:
You're a legend Ken

Issue #361, opened 2020-07-03
Support for the dedicated release branch flow


ArazAbishov:
# Challenges

The current bumping flow of the Beachball introduces a few challenges:
- Potential inconsistency between versions on the master, and artifacts published to NPM.
- Code reverts require manually created change files
  
## Potential inconsistency
The process of version bumping in Beachball is not atomic, creating the possibility of inconsistency between versioning on master/dev branches and artifacts published to NPM. In relatively small repositories that are quick to build are not as prone to this issue as larger mono repositories that take more than an hour to build.

For example, if a PR is submitted to master shortly after `beachball publish` is executed, the PR may be merged with the master before the version update from Beachball. This results in a state, where the artifacts published to NPM do not correspond to any commit on master.

<img width="1108" alt="Screenshot 2020-06-25 at 11 55 01" src="https://user-images.githubusercontent.com/4133693/86467790-57f85e80-bd36-11ea-8ce3-81fa07e9dbff.png">

# Proposal
Introduce a `master/dev` branch and a `release` branch. All PRs are supposed to target the `dev` branch. Beachball will merge changes from the `dev` branch to the `release` branch and update versions in the process. Versions are only maintained on the release branch, including the `changelog` files. Only one Beachball merge process can be executed at a time. Artifacts are built and released based on the release branch only.

kenotron:
## Steps
Here are steps in the release build job:

1. `git merge --no-commit master`
2. bump (changes changelog, package.json)
3. `git add .`
4. `git merge --continue`
5. do BUILD / TEST, whatever
6. beachball publish to npm, pushes package.json changes back into RELEASE branch

## Discussion Topics
- whether beachball should get in the business of helping you merge in changes from master into the release branch as outlined in steps 1-4

## Estimated Work (if yes on discussion point)
- create a merge + bump command 
- any publish command changes needed to push back to release rather than master

arazabishov:
Thanks for the response @kenotron . After taking a closer look at the implementation steps at Midgard together with @bweggersen, we have realized that a few intermediate implementation steps are missing in the proposal:

1. When merging dev/master into the rolling release branch, beachball needs to use only those change files that have been added since the last bump. This is necessary because dev/master is going to accumulate change files indefinitely. The API might look something like: `beachball bump --since <REF>`.
2. Ability to run `publish` without executing `bump` and revert, i.e. steps 1, 2, and 8 from the [publish documentation](https://microsoft.github.io/beachball/publish).

PR #373, opened 2020-08-03
--since flag implementation for bump and publish commands

arazabishov:
This PR includes implementation of the `--since` flag that allows users to specify a range of change files used by the `bump` and `publish` commands. The value passed to the `--since` flag can be any git ref, i.e. commit SHA, branch name or tag. 

**Behaviour of `bump --since <ref>`:** with `bump --since C1` executed for the repo containing `C1 <- C2 <- C3` commits, `bump` is going to pick-up change files introduced in commits C2 and C3. Once versions are bumped, corresponding change files will be deleted.

**Behaviour of `publish --since <ref>`:** similar to `bump --since <ref`, with the difference that publish is going to upload affected packages to an npm registry.

The flag does not affect the behaviour of `check`, `change` or `sync` commands. This PR implements part of a
solution for the issue https://github.com/microsoft/beachball/issues/361.

### DIFF
--- a/packages/beachball/src/git/index.ts
+++ b/packages/beachball/src/git/index.ts
@@ -106,8 +106,8 @@ function processGitOutput(output: ProcessOutput) {
     return [];
   }
 
-  let changes = output.stdout;
-  let lines = changes.split(/\n/) || [];
+  let stdout = output.stdout;
+  let lines = stdout.split(/\n/) || [];
 
   return lines
     .filter(line => line.trim() !== '')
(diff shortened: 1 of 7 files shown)


packages/beachball/src/git/index.ts
```diff
@@ -81,25 +87,34 @@ export function fetchRemote(remote: string, remoteBranch: string, cwd: string) {
...
     console.error('Cannot gather information about changes: ', e.message);
   }
 }
 
+function processGitOutput(output: ProcessOutput) {
+  if (!output.success) {
+    return [];
+  }
+
+  let changes = output.stdout;
```
kenotron, line 109: nit: I think 'change' should be something else here

arazabishov: Good catch! I've also noticed that `getStagedChanges` could benefit from using `processGitOutput` and updated it :)

PR #376, opened 2020-08-06
`--no-bump` flag implementation

arazabishov:
`--no-bump` allows to skip the bump step when invoking the `publish` command. As a side effect, it also implies that publish is not going to have anything to commit or push to the remote repo. Hence, when using `--no-bump` you also get the `--no-push` behaviour. The PR makes necessary changes to the console output and documentation to make that clear to the user.


sunilsurana:
I guess this wiki needs to be updated with this option . https://microsoft.github.io/beachball/cli/options.html#bumping-and-publishing-options

PR #403, opened 2020-09-10
Update error logic for disallowedChangeTypes in groups option

xugao:
Only throw error when disallowedChangeType is defined in package option and not defined in groups option

### DIFF
--- a/src/validation/isValidGroupOptions.ts
+++ b/src/validation/isValidGroupOptions.ts
@@ -20,7 +20,7 @@ export function isValidGroupOptions(root: string, groups: VersionGroupOptions[])
   for (const grp of Object.keys(packageGroups)) {
     const pkgs = packageGroups[grp].packageNames;
     for (const pkgName of pkgs) {
-      if (packageInfos[pkgName].options.disallowedChangeTypes) {
+      if (packageInfos[pkgName].options.disallowedChangeTypes && !packageGroups[grp].disallowedChangeTypes) {
         console.error(
           `Cannot have a disallowedChangeType inside a package config (${pkgName}) when there is a group defined; use the groups.disallowedChangeTypes instead.`
         );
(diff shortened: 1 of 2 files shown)


src/validation/isValidGroupOptions.ts
```diff
@@ -20,7 +20,7 @@ export function isValidGroupOptions(root: string, groups: VersionGroupOptions[])
   for (const grp of Object.keys(packageGroups)) {
     const pkgs = packageGroups[grp].packageNames;
     for (const pkgName of pkgs) {
-      if (packageInfos[pkgName].options.disallowedChangeTypes) {
+      if (packageInfos[pkgName].options.disallowedChangeTypes && !packageGroups[grp].disallowedChangeTypes) {
```
ecraig12345, line 23: Not quite understanding what's going on here, sorry--would you mind explaining a bit more?

xugao: @kenotron added this validation originally, so please correct me if i'm wrong here. 

My understanding is we want users to define `disallowedChangeTypes ` for the entire group rather than allowing each package to define individually. otherwise beachball doesn't know how to behave if packages within the group have different `disallowedChangeTypes`.

ecraig12345 (dismissed):

ecraig12345 (approved):
I approved but then undid it and now am re-approving... 😆  I'm still having trouble wrapping my mind around what each condition means, so I think it would be better to get Ken to review. But also it's not hugely important and we can always fix it later if it turns out to be wrong.

src/validation/isValidGroupOptions.ts
kenotron: The intention is to disallow any "disallowedChangeTypes" settings within a package for any packages defined in a group. This is regardless of whether you have a setting for the group or not. I'm not sure what bug you're addressing. Would like to see an example here.

xugao:
fix is made in #405

PR #405, opened 2020-09-11
Distinguish between combinedOptions vs packageOptions so disallowChangeType check can be done

kenotron:
In a version group, the disallowChangeType must not be specified at the package level. However, the packageInfo options has root options mixed into it. This made the validation of the option to be overly strict. This change preserves which options were set by the package itself. This will allow validations to be more precise.

### DIFF
--- a/src/monorepo/infoFromPackageJson.ts
+++ b/src/monorepo/infoFromPackageJson.ts
@@ -1,8 +1,9 @@
 import path from 'path';
 import { PackageInfo, PackageJson } from '../types/PackageInfo';
-import { getPackageOptions } from '../options/getPackageOptions';
+import { getActualPackageOptions, getPackageOptions } from '../options/getPackageOptions';
 
 export function infoFromPackageJson(packageJson: PackageJson, packageJsonPath: string): PackageInfo {
+  const actualOptions = getActualPackageOptions(path.dirname(packageJsonPath));
   return {
     name: packageJson.name!,
     version: packageJson.version,
@@ -11,6 +12,7 @@ export function infoFromPackageJson(packageJson: PackageJson, packageJsonPath: s
     devDependencies: packageJson.devDependencies,
     peerDependencies: packageJson.peerDependencies,
     private: packageJson.private !== undefined ? packageJson.private : false,
-    options: getPackageOptions(path.dirname(packageJsonPath)),
+    combinedOptions: getPackageOptions(actualOptions),
+    packageOptions: getActualPackageOptions(path.dirname(packageJsonPath)),
   };
 }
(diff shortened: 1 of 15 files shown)


src/types/PackageInfo.ts
```diff
@@ -24,7 +24,8 @@ export interface PackageInfo {
   devDependencies?: PackageDeps;
   peerDependencies?: PackageDeps;
   private: boolean;
-  options: PackageOptions;
+  combinedOptions: PackageOptions;
+  packageOptions: PackageOptions;
```
xugao, line 28: nit: would be good to have comments for these here

xugao (approved):

xugao (changes requested):
packageOptions is returning the root config for some reason in FUI. looking at it more ...

src/options/getPackageOptions.ts
```diff
@@ -6,16 +6,34 @@ import { getDefaultOptions } from './getDefaultOptions';
...
+  try {
+    const results = configExplorer.load(packagePath);
+    return results && results.config;
+  } catch (e) {
+    // File does not exist, returns the default packageOptions
+    return {
+      gitTags: defaultOptions.gitTags,
+      disallowedChangeTypes: defaultOptions.disallowedChangeTypes,
+      defaultNpmTag: defaultOptions.defaultNpmTag,
+    };
```
xugao, line 37: this is not quite right; they will overwrite root config when we combine:
```
    ...rootOptions,
    ...actualPackageOptions,
```

xugao: i think we should return `Partial<PackageOptions>` type and `{}` value from  `getPackageOption`

src/monorepo/infoFromPackageJson.ts
```diff
@@ -11,6 +12,7 @@ export function infoFromPackageJson(packageJson: PackageJson, packageJsonPath: s
     devDependencies: packageJson.devDependencies,
     peerDependencies: packageJson.peerDependencies,
     private: packageJson.private !== undefined ? packageJson.private : false,
-    options: getPackageOptions(path.dirname(packageJsonPath)),
+    combinedOptions: getCombinedPackageOptions(actualOptions),
+    packageOptions: getPackageOptions(path.dirname(packageJsonPath)),
```
xugao, line 16:
```suggestion
    packageOptions: actualOptions,
```

src/options/getPackageOptions.ts
```diff
@@ -6,16 +6,29 @@ import { getDefaultOptions } from './getDefaultOptions';
...
 }
+
+/**
+ * Gets all the package options from the configuration file of the package itself without inheritance
+ */
+export function getPackageOptions(packagePath: string): Partial<PackageOptions> {
+  const configExplorer = cosmiconfigSync('beachball', { cache: false });
+  try {
+    const results = configExplorer.load(packagePath);
+    return results && results.config;
```
xugao, line 29:
```suggestion
    const results = configExplorer.load(path.join(packagePath, 'package.json'));
    return (results && results.config) || {};
```

src/types/PackageInfo.ts
kenotron: yuppers!

xugao (approved):

PR #422, opened 2020-10-19
Update sync command to respect the --tag value

arazabishov:
Some scenarios require an ability to choose a version tag for syncing packages. By default, beachball uses `latest` and does not allow to choose the distribution tag. This PR adds support for `--tag` in the sync command, as well as adds a couple of E2E tests :)

**cc:** @VincentBailly, @bweggersen


kenotron:
Thanks! This is well thought out and tested code 👍

PR #448, opened 2020-12-01
Expose disallowed-change-types as a CLI option

arazabishov:
The `disallowed-change-types` option allows the user to specify the `disallowedChangeTypes` as CLI args without adding the `beachball` config. It is especially useful on CI, where different policies can specify `disallowedChangeTypes` without changing the beachball config on the disk.

### DIFF
--- a/change/beachball-d030ff29-357f-4404-be5b-0b800e7582da.json
+++ b/change/beachball-d030ff29-357f-4404-be5b-0b800e7582da.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Expose 'disallowed-change-types' as a CLI option",
+  "packageName": "beachball",
+  "email": "arabisho@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


kenotron (approved):

NickGerleman:
We absorbed this into our repo, and it seems to have broken disallowedChangeTypes in package.json, and more concerningly led to one of our packages seeing a patch version bump with only prerelease change files. Can isolate to the latest Beachball version since we can see the bump change happen for Verdaccio runs before and after.

arazabishov:
I'm sorry to hear that @NickGerleman, I will investigate this. Can you provide a simple example where I can repro this bug?

NickGerleman:
https://github.com/microsoft/react-native-windows/blob/master/vnext/package.json is an example of a package with disallowedChnageType in the package.json. After the change, users were able to select disallowed options, and they seemed not to be propagated to beachball when bumping from dependant packages.

arazabishov:
@NickGerleman Thanks for reporting this. The fix is in review.

arazabishov:
The fix is merged now :)

PR #452, opened 2020-12-08
Parameterise dependentChangeType and fix a bug of the --type arg being ineffective

arazabishov:
This PR:
 * Fixes a bug: when the `--type` argument is passed without `--message`, the `response.type` always ends-up being `undefined`. This happens because the list of questions does not include the change type prompt, anticipating the value to be available from `options.type`, but the `response.type` receives undefined as the message prompt overrides it. 
 * Adds a new CLI argument `--dependent-change-type` that allows overriding the default change type calculated for change files. This flag is handy when the PR author knows what kind of change (other than `none` or `patch`) is anticipated for dependent packages. For example, if package C in graph A <- B <- C receives a new pre-release version 1.0.0-2, I would want B and C also to receive a pre-release bump.

### DIFF
--- a/change/beachball-b0d9adf7-3d09-4bf0-92b4-9a1d6b5a55fc.json
+++ b/change/beachball-b0d9adf7-3d09-4bf0-92b4-9a1d6b5a55fc.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Expose dependent-change-type as CLI argument",
+  "packageName": "beachball",
+  "email": "arabisho@microsoft.com",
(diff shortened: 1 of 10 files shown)


christiango:
Nit: update the PR description and title to have consistent spelling of dependent. It says dependant in the description, but dependent in code. I think dependent is the way we'd want to spell it, so I'm just asking to update the PR description and title

christiango:
Can the description be updated to provide more context on the usage of --dependent-change-type?

christiango (approved):

arazabishov:
Fixed the typos and updated the description!

PR #453, opened 2020-12-08
Implement the 'disallow-deleted-change-files' flag

arazabishov:
When `beachball` is invoked together with the `--disallow-deleted-change-files` flag, beachball will verify that no change files were deleted by inspecting the diff between head and `--branch`.

### DIFF
--- a/src/help.ts
+++ b/src/help.ts
@@ -25,24 +25,25 @@ Commands:
 
 Options:
 
-  --registry, -r          - registry, defaults to https://registry.npmjs.org
-  --tag, -t               - for the publish command: dist-tag for npm publishes
-                          - for the sync command: will use specified tag to set the version
-  --branch, -b            - target branch from origin (default: master)
-  --message, -m           - for the publish command: custom publish message for the checkin (default: applying package updates);
-                            for the change command: description of the change
-  --no-push               - skip pushing changes back to git remote origin
-  --no-publish            - skip publishing to the npm registry
-  --no-bump               - skip both bumping versions and pushing changes back to git remote origin when publishing;
-  --help, -?, -h          - this very help message
-  --yes, -y               - skips the prompts for publish
-  --package, -p           - manually specify a package to create a change file; creates a change file regardless of diffs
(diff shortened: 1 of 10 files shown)


src/help.ts
```diff
@@ -41,6 +41,7 @@ Options:
   --since             - for the bump command: allows to specify the range of change files used to bump packages by using git refs (branch name, commit SHA, etc);
                         for the publish command: bumps and publishes packages based on the specified range of the change files.
   --keep-change-files - for the bump and publish commands: when specified, both bump and publish commands do not delete the change files on the disk.
+                        for the check command: verifies that no change files were deleted between head and target branch.
```
christiango, line 44: It feels a bit strange to have this same CLI option do two very different things. Also, should this be opt out instead of opt-in?

src/validation/validate.ts
```diff
@@ -55,6 +56,11 @@ export function validate(options: BeachballOptions, validateOptionsOverride?: Pa
       console.log(options.changehint);
       process.exit(1);
     }
+
+    if (options.keepChangeFiles && areChangeFilesDeleted(options)) {
+      console.error('ERROR: Change files must not be deleted!');
```
christiango, line 61: Can we output what files were deleted? Or does that already happen in the areChangeFilesDeleted flow?

arazabishov: No, deleted files are not printed anywhere. I can fix that quickly.

src/help.ts
arazabishov: I'm open to adding a new flag. Do you think `--no-change-files-deleted` suits this purpose better? I think this flag should stay opt-in, otherwise this change won't be backwards compatible with the existing behavior of beachball.

christiango: Yeah, I like no change files deleted!

arazabishov: I've switched to `disallow-deleted-change-files` because `yargs` logic for the negation of booleans prefixed with `no-*` complicates everything.

christiango (approved):

christiango (approved):

PR #465, opened 2020-12-23
Remove package-lock.json

bweggersen:
yarn.lock is the most updated lock-file, so we should remove the package-lock.json to avoid confusion.


ecraig12345:
Sorry nobody saw this sooner. Could you also add `package-lock.json` to `.gitignore` to ensure it doesn't get introduced again?

ecraig12345:
I removed the file and added it to gitignore in another change. Thanks again for the PR though!

PR #466, opened 2020-12-24
Add support for multiple project roots within a repo

### LINKED ISSUE #439: beachball broken when workspace-root is not the same as the git-root
I'm running into issues with beachball when the workspace-root (package-root?) doesn't match the git-root.  Specifically, when package.json is not in the root of the repo, but instead a subdirectory.  Scenarios like this were fixed recently for workspace-tools (microsoft/workspace-tools#18), but it looks like it needs to be fixed here too.

https://github.com/microsoft/beachball/blob/3186f8ed09abbbad83e9cae2f4188ecc1b826b9d/src/git/index.ts#L369

bweggersen:
Instead of first looking for the git root, beachball will now find the closest project root to cwd. Through workspace-tools beachball supports yarn, rush, and pnpm workspaces. If none of these workspaces can be found, beachball will default to using the git root.

The most important change is changing the function `findGitRoot` to:

```ts
export function findProjectRoot(cwd: string) {
  let workspaceRoot: string | undefined;
  try {
    workspaceRoot = getWorkspaceRoot(cwd);
  } catch {}

  return workspaceRoot || searchUp('.git', cwd);
}
```

Resolves #439


src/fixtures/multi-monorepo.ts
```diff
@@ -0,0 +1,107 @@
+import * as process from 'process';
```
arazabishov, line 1: Nit: is it common to use kebab-case for ts files? :) The rest of the project files are using camel case.

src/paths.ts
```diff
@@ -1,5 +1,6 @@
 import path from 'path';
 import fs from 'fs-extra';
+import { getWorkspaceRoot } from 'workspace-tools';
```
arazabishov, line 3: if we need only this function, should we use `find-yarn-workspace-root` instead? That's what `workspace-tools` is using under the hood.

package.json
```diff
@@ -43,8 +43,9 @@
     "prompts": "~2.1.0",
     "semver": "^6.1.1",
     "toposort": "^2.0.2",
-    "yargs-parser": "^20.2.4",
-    "uuid": "^8.3.1"
+    "uuid": "^8.3.1",
+    "workspace-tools": "^0.10.2",
```
arazabishov, line 47: As mentioned in the other comment, do we want to take a dependency on `workspace-tools` as a whole or just `find-yarn-workspace-root`?

src/paths.ts
```diff
@@ -25,16 +26,21 @@ export function searchUp(pathName: string, cwd: string) {
   return null;
 }
 
-export function findGitRoot(cwd: string) {
-  return searchUp('.git', cwd);
+export function findProjectRoot(cwd: string) {
+  let workspaceRoot: string | undefined;
+  try {
+    workspaceRoot = getWorkspaceRoot(cwd);
```
kenotron, line 32: I think what we need to do is:

1. leave findGitRoot() as-is because it is used as a validation check (isGitAvailable)
2. simply replace the majority of the usage of `findGitRoot(cwd)` with `getWorkspaceRoot(cwd)`

src/paths.ts
```diff
@@ -25,16 +26,21 @@ export function searchUp(pathName: string, cwd: string) {
...
+  return workspaceRoot || searchUp('.git', cwd);
 }
 
 export function findPackageRoot(cwd: string) {
   return searchUp('package.json', cwd);
 }
 
 export function getChangePath(cwd: string) {
-  const gitRoot = findGitRoot(cwd);
+  const gitRoot = findProjectRoot(cwd);
```
kenotron, line 43: These lines are syntactically correct, but the meaning is incorrect in the variable names now...

```suggestion
  const root = getWorkspaceRoot(cwd);
```

Of course then you'll have more lines to fix below...

src/validation/areChangeFilesDeleted.ts
```diff
@@ -1,11 +1,11 @@
-import { findGitRoot, getChangePath } from '../paths';
+import { findProjectRoot, getChangePath } from '../paths';
 import { BeachballOptions } from '../types/BeachballOptions';
 import { getChangesBetweenRefs } from '../git';
 
 export function areChangeFilesDeleted(options: BeachballOptions): boolean {
   const { branch, path: cwd } = options;
 
-  const gitRoot = findGitRoot(cwd);
+  const gitRoot = findProjectRoot(cwd);
```
kenotron, line 8: same as the other suggestion: gitRoot -> root; findProjectRoot -> getWorkspaceRoot

src/validation/isGitAvailable.ts
```diff
@@ -1,8 +1,8 @@
-import { findGitRoot } from '../paths';
+import { findProjectRoot } from '../paths';
 import { git } from '../git';
 
 export function isGitAvailable(cwd: string) {
   const result = git(['--version']);
-  const gitRoot = findGitRoot(cwd);
+  const gitRoot = findProjectRoot(cwd);
```
kenotron, line 6: This should still be "findGitRoot" because the function here is trying to verify that .git exists:

```suggestion
  const gitRoot = findGitRoot(cwd);
```

kenotron (changes requested):
1. Araz's nit is spot on with the naming of files.
2. Preserve findGitRoot() for the validator
3. Straight up replace usage of "findGitRoot()" in other cases with "getWorkspaceRoot" from workspace-tools

src/paths.ts
kenotron: I think we should use `workspace-tools` wholesale, much of the util code in beachball had been ported over there and has become a sort of dupe.

src/fixtures/multi-monorepo.ts
bweggersen: Maybe I was just hungry?! 😛 Thanks for pointing it out!

src/paths.ts
bweggersen: I think I understand what Ken is saying (wholesale?), that we should use `workspace-tools`, and I agree with that. `workspace-tools` also supports rush and pnpm workspaces.

src/paths.ts
bweggersen: If I replace `findGitRoot(cwd)` with `getWorkspaceRoot(cwd)`, then we loose the fallback to git root if there's no workspace in the project. Should I rather:
1. Bring back `findGitRoot()` and use it in `isGitAvailable`
2. Keep `findProjectRoot()` as it is in the PR to get the fallback functionality

src/fixtures/multi-monorepo.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      fs.mkdirpSync(path.join(tmpRepo.rootPath, pkg));
+
+      fs.writeJSONSync(path.join(tmpRepo.rootPath, packageJsonFile), packageJsonFixture, {
+        spaces: 2,
+      });
+      tmpRepo.commitChange(packageJsonFile);
+    }
+
+    tmpRepo.commitChange('package.json', JSON.stringify({ name: 'multi-monorepo-fixture', version: '1.0.0' }, null, 2));
```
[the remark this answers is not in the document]
bweggersen: The package.json is not strictly needed. I can remove it. The yarn.lock is just to inform `workspace-tools` that this is a `yarn` project (see [code in workspace-tools](https://github.com/microsoft/workspace-tools/blob/cd2c09d2fbfc1127f85781d5243bf7b21f4116d1/src/workspaces/implementations/index.ts#L8)). I could also add a yarn.lock file to each sub-project.

src/paths.ts
bweggersen: I changed the variable name, but kept the function name since I want to use git root as fallback.

src/validation/isGitAvailable.ts
bweggersen: Thanks! Good catch 😅

src/paths.ts
bweggersen: Due to time zone differences, I'll just update the branch with my suggestions and then you can approve or comment on that.

kenotron: That's fair. Let's just go with what you have.

kenotron (approved):

PR #471, opened 2021-01-05
Rolling-up the changelogs to dependents

arazabishov:
Currently, beachball generates changelogs only for packages that are described in change files. Dependent packages, however, do not receive new changelog entries, even though their versions get bumped. 

This PR enables beachball to rollup a changelog entry to dependents, such as: `Bump ${package} to v${version}` making it obvious what has caused the version bump.

### DIFF
--- a/src/bump/updateRelatedChangeType.ts
+++ b/src/bump/updateRelatedChangeType.ts
@@ -1,4 +1,4 @@
-import { getMaxChangeType, MinChangeType } from '../changefile/getPackageChangeTypes';
+import { getMaxChangeType, MinChangeType, updateChangeInfoWithMaxType } from '../changefile/getPackageChangeTypes';
 import { BumpInfo } from '../types/BumpInfo';
 import { ChangeInfo } from '../types/ChangeInfo';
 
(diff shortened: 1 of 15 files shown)


src/bump/updateRelatedChangeType.ts
```diff
@@ -9,50 +9,80 @@ import { BumpInfo } from '../types/BumpInfo';
...
+  dependentChangeInfos: Map<string, Array<ChangeInfo>>,
   bumpDeps: boolean
 ) {
   const { packageChangeTypes, packageGroups, dependents, packageInfos, dependentChangeTypes, groupOptions } = bumpInfo;
 
-  const disallowedChangeTypes = packageInfos[pkgName].combinedOptions?.disallowedChangeTypes ?? [];
+  const packageInfo = packageInfos[pkgName];
+  const disallowedChangeTypes = packageInfo.combinedOptions?.disallowedChangeTypes ?? [];
+
+  let depChangeType = {
```
kenotron, line 22: maybe cleaner to have getMaxChangeType change its purpose to "updateChangeInfoWithMaxType"

src/bump/updateRelatedChangeType.ts
```diff
@@ -9,50 +9,80 @@ import { BumpInfo } from '../types/BumpInfo';
  */
 export function updateRelatedChangeType(
   pkgName: string,
-  changeType: ChangeType,
+  changeType: ChangeInfo,
```
kenotron, line 12: changeType really is "changeInfo" now

PR #472, opened 2021-01-07
Prevent prompt from overriding the message passed by CLI args

arazabishov:


src/changefile/promptForChange.ts
```diff
@@ -98,6 +98,11 @@ export async function promptForChange(options: BeachballOptions) {
         response = { ...response, type: options.type };
       }
 
+      // fallback to the options.message if message is absent in the user input
+      if (!response.comment && options.message) {
+        response = { ...response, comment: options.message };
```
bweggersen, line 103: This value gets set on line 85. I assume that something might happen on line 89 that causes us to have to set comment to `option.message` again. Can you explain this scenario?

arazabishov: Sure. Code at line 89 is responsible for prompting the user input (change type and message). When one of those values are passed through CLI (for example message), prompt will be missing corresponding question. If question is missing, response will have an undefined value for comment, effectively overriding what user has passed through CLI.

bweggersen (approved):

PR #473, opened 2021-01-13
Remove the double negative in patch change hint 

eivisand:
The current hints for patch and minor changes use different language to tell users that they should only have backwards compatible changes. This change unifies the language, going for the least complicated phrasing.

### DIFF
--- a/src/changefile/promptForChange.ts
+++ b/src/changefile/promptForChange.ts
@@ -36,7 +36,7 @@ export async function promptForChange(options: BeachballOptions) {
       message: 'Change type',
       choices: [
         ...(showPrereleaseOption ? [{ value: 'prerelease', title: ' [1mPrerelease[22m - bump prerelease version' }] : []),
-        { value: 'patch', title: ' [1mPatch[22m      - bug fixes; no backwards incompatible changes.' },
+        { value: 'patch', title: ' [1mPatch[22m      - bug fixes; backwards compatible changes.' },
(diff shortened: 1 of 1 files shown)


src/changefile/promptForChange.ts
```diff
@@ -36,7 +36,7 @@ export async function promptForChange(options: BeachballOptions) {
       message: 'Change type',
       choices: [
         ...(showPrereleaseOption ? [{ value: 'prerelease', title: ' [1mPrerelease[22m - bump prerelease version' }] : []),
-        { value: 'patch', title: ' [1mPatch[22m      - bug fixes; no backwards incompatible changes.' },
+        { value: 'patch', title: ' [1mPatch[22m      - bug fixes; backwards compatible changes.' },
         { value: 'minor', title: ' [1mMinor[22m      - small feature; backwards compatible changes.' },
```
ecraig12345, line 40: Thanks for the contribution but now that you point this out, I actually think both the patch and minor descriptions are unclear. I'd suggest something like this instead:
```suggestion
        { value: 'patch', title: ' [1mPatch[22m      - bug fixes; no API changes.' },
        { value: 'minor', title: ' [1mMinor[22m      - small feature; backwards compatible API changes.' },
```
@xugao and/or @kenotron what do you think?

xugao: Agreed!

ecraig12345 (approved):

PR #474, opened 2021-01-14
Fix setting response.type in promptForChange

kubkon:
This commit handles edge case for properly setting `response.type` in `promptForChange` in a situation when `options.type` is `undefined` or `null` **and** there is only one viable default `ChangeType` option (`'none'`) in which case the type prompt won't be shown, and we fall back to `'none'` by default.

Prior to this change, since `options.type` would be left `undefined` and the type prompt not shown, the CLI would incorrectly exit with "Prompt response contains invalid change type" without any other hint to the user.

For more context, this problem emerged when we were adding a new package to the [microsoft/fluentui](https://github.com/microsoft/fluentui) repo: microsoft/fluentui#16458.

---
EDIT:

After offline discussion with @ecraig12345, we agreed it would make sense to emit a warning if the user selects `none` as the change type, or ask for confirmation. I went a step further and figure it would make sense to actually confirm with the user their choices regardless. Therefore, this PR also adds a snippet which prints out a confirmation dialog to confirm their choices. Example output after this change looks like:

```
Checking for changes against "origin/master"
fetching latest from remotes "origin/master"
Found changes in the following packages:
  @fluentui/cra-template
fetching latest from remotes "origin/master"

Please describe the changes for: @fluentui/cra-template
✔ Describe changes (type or choose one) › feat(cra-template): boot up CRA template
Your response:
{
  changeType: "none",
  comment:    "feat(cra-template): boot up CRA template",
}
✔ Do you want to continue? … no
Aborting...
```

Hopefully, this should shield the user a little bit better from making unwanted decision, for instance, like in the snippet above, selecting `'none'` as the change type which is undesired and should be investigated and fixed before running `beachball` against the changes.

### DIFF
--- a/src/changefile/promptForChange.ts
+++ b/src/changefile/promptForChange.ts
@@ -101,6 +101,9 @@ export async function promptForChange(options: BeachballOptions) {
       //    but we didn't display it due to showChangeTypePrompt === false;
       //    so set the type to 'none'
       if (!response.type) {
+        if (!options.type) {
+          console.log("WARN: change type 'none' assumed by default");
+        }
         response = { ...response, type: options.type || 'none' };
       }
 
--- a/src/changefile/promptForChange.ts
+++ b/src/changefile/promptForChange.ts
@@ -93,9 +93,15 @@ export async function promptForChange(options: BeachballOptions) {
         return;
       }
 
-      // fallback to the options.type if type is absent in the user input
-      if (!response.type && options.type) {
-        response = { ...response, type: options.type };
+      // if type is absent in the user input, there are two possiblities for
+      // proceeding next:
+      // 1) if options.type is defined, use that
+      // 2) otherwise, we hit the edge case when options.type is undefined
+      //    and there was only one possible ChangeType to display, 'none'
+      //    but we didn't display it due to showChangeTypePrompt === false;
+      //    so set the type to 'none'
+      if (!response.type) {
+        response = { ...response, type: options.type || 'none' };
       }
 
       // fallback to the options.message if message is absent in the user input


kubkon:
cc @kenotron @ecraig12345 @Hotell

src/changefile/promptForChange.ts
```diff
@@ -93,9 +93,15 @@ export async function promptForChange(options: BeachballOptions) {
...
       }
 
-      // fallback to the options.type if type is absent in the user input
-      if (!response.type && options.type) {
-        response = { ...response, type: options.type };
+      // if type is absent in the user input, there are two possiblities for
+      // proceeding next:
+      // 1) if options.type is defined, use that
+      // 2) otherwise, we hit the edge case when options.type is undefined
+      //    and there was only one possible ChangeType to display, 'none'
```
ecraig12345, line 100: Thanks for figuring this out! A clarifying question: I thought in the specific case you encountered, there should have been two possible change types (none and prerelease)? Or am I misunderstanding?

kubkon: Hmm, interesting! There actually was only `'none'` as an option available. If there were two options to choose from, the bug would not surface. Lemme quickly have a look why `'prelease'` was not available as an option in this situation.

kubkon: Hmm, ok, so `semver.prelease` returns `null` since the package version is just a triple, and it expects a quadruple, e.g., `0.1.0-pre`. We can continue the conversation going in the PR proper https://github.com/microsoft/fluentui/pull/16458.

ecraig12345: Hmm I think you've found an interesting edge case which we might want @kenotron's input on how to handle.

In Fluent UI right now we have a repo-wide config with `disallowedChangeTypes: ['major', 'minor', 'patch']`. This implicitly assumes that all published packages have a prerelease semver. For the package that had changes, that wasn't the case, which is why beachball only offered `none` as an option.

Your change here is one possible way to better handle that case, but I wonder if an even better way would be to have beachball show an error? (suggesting to either add a prerelease to the package's version, or add a package-specific override to allow more change types)

kubkon: If you look closely, this is not an erroneous condition, and shouldn't be treated as such. First of all, in [promptForChange.ts#L49](https://github.com/microsoft/beachball/blob/f7f6170068a9e8548c9bff62f05c316357b65918/src/changefile/promptForChange.ts#L49) we ensure that we have at least one valid option available. Then, in [promptForChange.ts#L68](https://github.com/microsoft/beachball/blob/f7f6170068a9e8548c9bff62f05c316357b65918/src/changefile/promptForChange.ts#L68), we ensure that we only show the `ChangeType` prompt if we have at least 2 valid choices, meaning, the case when both `!options.type` evaluates to `true`, and having one option available, should either be shown to the end-user, or should override `options.type` with that option. This is in fact already enforced for situations when `!options.type` evaluates to `true`, and there is only one valid `ChangeType` option available, and `!option.message` also evaluates to `true`. In this case, we completely bypass the prompts in [promptForChange.ts#88](https://github.com/microsoft/beachball/blob/f7f6170068a9e8548c9bff62f05c316357b65918/src/changefile/promptForChange.ts#L88), and set the `response` to the defaults in [promptForChange.ts#L83](https://github.com/microsoft/beachball/blob/f7f6170068a9e8548c9bff62f05c316357b65918/src/changefile/promptForChange.ts#L83) where `type := 'none'`.

ecraig12345:
>  I went a step further and figure it would make sense to actually confirm with the user their choices regardless. Therefore, this PR also adds a snippet which prints out a confirmation dialog to confirm their choices.

Sorry to keep going back and forth on this but I actually think it would be very obnoxious to prompt for confirmation every time. I'd prefer to add a warning only for the specific case where `none` is inferred as the only valid change type.

kubkon:
> > I went a step further and figure it would make sense to actually confirm with the user their choices regardless. Therefore, this PR also adds a snippet which prints out a confirmation dialog to confirm their choices.
> 
> Sorry to keep going back and forth on this but I actually think it would be very obnoxious to prompt for confirmation every time. I'd prefer to add a warning only for the specific case where `none` is inferred as the only valid change type.

No probs at all! I actually thought about giving the CLI an option like `-y` for `apt-get` which would mean confirming is always implicit. But then, that would introduce complexity into the package, plus we still need some response from the user which would be somewhat clunky if `-y` would only skip confirmation. Anyhow, I'm happy to revert to just a warning 👍

kubkon:
@ecraig12345 done in 569388e!

ecraig12345 (approved):

PR #475, opened 2021-01-15
Allow the "type" CLI arg to accept multiple types

NickGerleman:
The "type" arg is useful when scripting Beachball to avoid prompting. There are sometimes cases where react-native-windows wants to generate changefiles of different types dependent on what the package allows. E.g. where one package only allows prerelease types, but others prefer patch.

This change allows the "type" argument to accept a list of types in order of preference, where the first allowed type is picked. This maintains compatibility with previous usage, since a single argument will be treated by `yargs-parser` as a single element array.

There is not great infrastructure present to test the change prompt, so validation was manual. I linked the development beachball bits to a separate repo and tested the cases of no type arg, one type arg, multiple type args (with 1 vs 2 winning).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ target branch from origin (default: master)
 
 ### --message, -m
 
-custom message for the checkin (default: applying package updates)
+- for the publish command: custom message for the checkin (default: applying package updates)
+- for the change command: uses the given changelog msessage instead of prompting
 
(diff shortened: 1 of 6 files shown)


NickGerleman:
Closing this in favor of a couple of other changes to allow some greater extensibility.

PR #477, opened 2021-01-15
Allow passing --config to specify config file

NickGerleman:
Automated testing for this change was a bit tricky. To make options code more testable, we allowing injecting argv instead of grabbing from the process, which we should not mutate, and may have flags we don't want. E.g. E2ETest will run jest with a `--config` flag we don't want in our tests.

This exposed some issues where `getCliOptions` was accidentally trimming bits off of process.argv, and we needed to change the CLI options cache to allow different values of argv to have different results. We do this by special casing the cache to only cache when the input is process.argv, which should be immutable and will be the only option used in production code.

### DIFF
--- a/src/options/getCliOptions.ts
+++ b/src/options/getCliOptions.ts
@@ -3,21 +3,31 @@ import { CliOptions } from '../types/BeachballOptions';
 import { findProjectRoot } from '../paths';
 import { getDefaultRemoteBranch } from '../git';
 
-// CLI Options cache
-let cliOptions: CliOptions;
+let cachedCliOptions: CliOptions;
 
-export function getCliOptions(): CliOptions {
-  if (cliOptions) {
-    return cliOptions;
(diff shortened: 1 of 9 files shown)


src/options/getCliOptions.ts
```diff
@@ -3,21 +3,31 @@ import { CliOptions } from '../types/BeachballOptions';
...
 import { getDefaultRemoteBranch } from '../git';
 
-// CLI Options cache
-let cliOptions: CliOptions;
+let cachedCliOptions: CliOptions;
 
-export function getCliOptions(): CliOptions {
-  if (cliOptions) {
-    return cliOptions;
+export function getCliOptions(argv: string[]) : CliOptions {
```
kenotron, line 8: nice change! makes it more easily testable.

kenotron (approved):

PR #481, opened 2021-01-28
feat(change): add --no-commit flag to CLI

kubkon:
This commit adds `--no-commit` flag which only affects `beachball change` command. With this flag set, the change files generated with `beachball change` will only be staged for commit in git, but the commit will have to be done manually by the caller of `beachball`. The defaults stay the same, i.e., by default we still stage and autocommit the generated change files for the user with message 'Change file'.

Suggested usage for only staging change files:

```
beachball change --no-commit
```

This addition is motivated by a few of us working on [microsoft/fluentui](https://github.com/microsoft/fluentui) constantly finding ourselves uncommitting the change files and then squashing them with the actual changes to the codebase.

### DIFF
--- a/src/__e2e__/change.test.ts
+++ b/src/__e2e__/change.test.ts
@@ -1,0 +1,82 @@
+import fs from 'fs-extra';
+import { RepositoryFactory } from '../fixtures/repository';
+import { git } from '../git';
+import { change } from '../commands/change';
+import { BeachballOptions } from '../types/BeachballOptions';
+import { getChangePath } from '../paths';
+
+describe('change command', () => {
+  let repositoryFactory: RepositoryFactory | undefined;
+
+  function getChangeFiles(cwd: string): string[] {
+    const changePath = getChangePath(cwd);
+    const changeFiles = changePath && fs.existsSync(changePath) ? fs.readdirSync(changePath) : [];
+    return changeFiles;
+  }
+
+  afterEach(async () => {
+    if (repositoryFactory) {
+      repositoryFactory.cleanUp();
+      repositoryFactory = undefined;
+    }
+  });
+
+  it('create change file but git stage only', async () => {
+    repositoryFactory = new RepositoryFactory();
+    await repositoryFactory.create();
+    const repo = await repositoryFactory.cloneRepository();
+
+    await repo.commitChange(
+      'packages/pkg-1/package.json',
+      JSON.stringify({
+        name: 'pkg-1',
+        version: '1.0.0',
+      })
+    );
+
+    await change({
+      type: 'minor',
+      dependentChangeType: 'patch',
+      package: 'pkg-1',
+      message: 'stage me please',
+      path: repo.rootPath,
+      commit: false,
+    } as BeachballOptions);
+
+    const output = git(['status', '-s'], { cwd: repo.rootPath });
+    expect(output.success).toBeTruthy();
+    expect(output.stdout.startsWith('A')).toBeTruthy();
+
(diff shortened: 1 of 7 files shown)


kubkon:
cc @ecraig12345 @Hotell

src/types/BeachballOptions.ts
```diff
@@ -38,6 +38,7 @@ export interface CliOptions {
   disallowDeletedChangeFiles?: boolean;
   prereleasePrefix?: string | null;
   configPath?: string;
+  commit: boolean;
```
Hotell, line 41: **suggestion:** I'd make this optional ->`commit?: boolean` ( default is true anyways ). With that you should be able to get rid of all those test changes where you've been forced to add `commit: true` by type checker

src/changefile/writeChangeFiles.ts
```diff
@@ -13,7 +13,8 @@ export function writeChangeFiles(
   changes: {
     [pkgname: string]: ChangeFileInfo;
   },
-  cwd: string
+  cwd: string,
+  commitChangeFiles?: boolean
```
Hotell, line 17: **suggestion:** you can use default args pattern thus change on line 41 can be removed

-> 
```diff
-commitChangeFiles?: boolean
+commitChanges = true  // TypeScript will properly infer that this argument is optional and type of boolean
```

↓

```
if (commitChanges) {
      commit('Change files', cwd);
}
```

Hotell (commented):
added 2 suggestions. Overall looks good! I'd love to see adding a test to cover this new behaviour.

src/types/BeachballOptions.ts
kubkon: I was on the fence with this one. Originally had it as optional, but then decided to go with a non-optional :D

src/changefile/writeChangeFiles.ts
kubkon: Excellent, thanks!

src/types/BeachballOptions.ts
Hotell: yeah, it is a little confusing - it's basically being "set" by that default argument that I mentioned, thus there is no reason having it as required. 

> thought: Not sure why this project doesn't use standard yargs for all the CLI orchestration, where this could have been handled in a "standard" way...

ecraig12345: > thought: Not sure why this project doesn't use standard yargs for all the CLI orchestration, where this could have been handled in a "standard" way...

You'd have to ask @kenotron to confirm but I think he switched from `yargs` to `yargs-parser` due to some issues in certain contexts related to statefulness of `yargs`? Or maybe just to be lighter weight, I forget.

ecraig12345 (approved):
Agreed with @Hotell that a test would be nice but otherwise LGTM, thanks!

kubkon:
> Agreed with @Hotell that a test would be nice but otherwise LGTM, thanks!

@Hotell and I have added a test for the change command in 54fb5fb.

src/__e2e__/change.test.ts
```diff
@@ -0,0 +1,82 @@
+import fs from 'fs-extra';
+import { RepositoryFactory } from '../fixtures/repository';
+import { git } from '../git';
+import { change } from '../commands/change';
+import { BeachballOptions } from '../types/BeachballOptions';
+import { getChangePath } from '../paths';
+
+describe.only('change command', () => {
```
Hotell, line 8: chore: remove `.only` pls

Hotell (approved):

PR #483, opened 2021-02-03
check: Add validation of published package deps

ecraig12345:
In Fluent UI we've repeatedly had issues where someone adds a new package which accidentally has `"private": true` set in its package.json, adds the new package as a dep of a published package, and this isn't detected until publishing fails.

Fix is to update `beachball check` to run the same validation logic that's used before publishing to catch these types of errors.

### DIFF
--- a/src/publish/validatePackageDependencies.ts
+++ b/src/publish/validatePackageDependencies.ts
@@ -9,7 +9,8 @@ export function validatePackageDependencies(bumpInfo: BumpInfo, quiet?: boolean)
   const { modifiedPackages, newPackages, packageInfos } = bumpInfo;
 
   const packagesToValidate = [...modifiedPackages, ...newPackages];
-  const allDeps: { [pkg: string]: string[] } = {};
+  /** Mapping from dep to all validated packages that depend on it */
+  const allDeps: { [dep: string]: string[] } = {};
   for (const pkg of packagesToValidate) {
     const { publish, reasonToSkip } = shouldPublishPackage(bumpInfo, pkg);
     if (!publish) {
--- a/src/publish/validatePackageDependencies.ts
+++ b/src/publish/validatePackageDependencies.ts
@@ -4,30 +4,38 @@ import { shouldPublishPackage } from './shouldPublishPackage';
 /**
  * Validate no private package is listed as package dependency for packages which will be published.
  */
-export function validatePackageDependencies(bumpInfo: BumpInfo): boolean {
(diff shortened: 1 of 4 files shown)


src/validation/validate.ts
```diff
@@ -95,6 +97,19 @@ export function validate(options: BeachballOptions, validateOptionsOverride?: Pa
     }
   }
 
+  if (!isChangeNeeded) {
+    const bumpInfo = gatherBumpInfo(options);
```
xugao, line 101: I think `gatherBumpInfo` will bump package versions in place (write to package.json). can you double check?

ecraig12345: I tried it out and it didn't cause any local changes.

ecraig12345: Would be ideal to also add a test for this whole scenario (including that it doesn't modify any files) but I'm not sure how hard that would be...

xugao: Ok, my bad - `performBump` is the one actually writes.

xugao (approved):
I'm good with this change. But pls have Ken to take a look as well

src/publish/validatePackageDependencies.ts
```diff
@@ -4,30 +4,39 @@ import { shouldPublishPackage } from './shouldPublishPackage';
...
-  packagesToValidate.forEach(pkg => {
+  /** Mapping from dep to all validated packages that depend on it */
+  const allDeps: { [dep: string]: string[] } = {};
+  for (const pkg of packagesToValidate) {
     const { publish, reasonToSkip } = shouldPublishPackage(bumpInfo, pkg);
     if (!publish) {
-      console.log(`Skipping package dep validation - ${reasonToSkip}`);
-      return;
+      if (!quiet) {
+        console.log(`Skipping package dep validation - ${reasonToSkip}`);
```
kenotron, line 18: So, I know this is an annoyance... but can we isolate the change in this PR without changing this function's signature?

I have in mind an overhaul of the logging of beachball to use something a lot like what's used inside `lage` where I can do a more flexible "reporter" infra (with a default that can have error levels).

kenotron (changes requested):
Can you please not change the validatePackageDeps signature for now? I can add a feature to turn these console.log's into a logger with levels that can be changed by cli options.

src/publish/validatePackageDependencies.ts
kenotron: #484 Is the issue I created for this kind of work.

kenotron (approved):

PR #486, opened 2021-02-05
Add postpublish hook

### LINKED ISSUE #482: Lockfile version not updated after publishing
Steps to reproduce:

- Bump a package version
- Release package version

`package.json` version is updated, but `package-lock.json` still has the old version

Expected behavior:
`package-lock.json` version is also updated (or `yarn.lock`)

Right now, I can run install to update the lockfile, but it would be great if beachball could update the version change automatically.

### LINKED ISSUE #488: Feature Request: add postpublish hook
Purpose: use this hook to modify files after publishing, or to kick off other processes. One concrete example is to update lock files, which currently don't get updated (using npm).

relates to: https://github.com/microsoft/beachball/issues/482

mmaclach:
Add postpublish hook used to modify files after publishing, or to kick off other processes. One concrete example is to update lock files, which currently don't get updated (using npm)--see #482.

Fixes #488 
Fixes #482

### DIFF
--- a/src/publish/publishToRegistry.ts
+++ b/src/publish/publishToRegistry.ts
@@ -43,15 +43,11 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
     return publish;
   });
 
-
-  // finally pass through doing the actual npm publish command
-  for (const pkg of packagesToPublish) {
-    const packageInfo = packageInfos[pkg];
-    console.log(`Publishing - ${packageInfo.name}@${packageInfo.version} with tag ${packageInfo.combinedOptions.tag}.`);
-
-    // if there is a prepublish hook perform a prepublish
-    const prepublishHook = options.hooks?.prepublish;
-    if (prepublishHook) {
+  // if there is a prepublish hook perform a prepublish pass, calling the routine on each package
+  const prepublishHook = options.hooks?.prepublish;
+  if (prepublishHook) {
+    for (const pkg of packagesToPublish) {
+      const packageInfo = bumpInfo.packageInfos[pkg];
       const maybeAwait = prepublishHook(
         path.dirname(packageInfo.packageJsonPath),
         packageInfo.name,
@@ -61,6 +57,12 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
         await maybeAwait;
       }
     }
+  }
+
+  // finally pass through doing the actual npm publish command
+  for (const pkg of packagesToPublish) {
+    const packageInfo = bumpInfo.packageInfos[pkg];
+    console.log(`Publishing - ${packageInfo.name}@${packageInfo.version} with tag ${packageInfo.combinedOptions.tag}.`);
 
     let result;
     let retries = 0;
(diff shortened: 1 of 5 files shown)


src/publish/publishToRegistry.ts
```diff
@@ -43,11 +43,15 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
...
   });
 
-  // if there is a prepublish hook perform a prepublish pass, calling the routine on each package
-  const prepublishHook = options.hooks?.prepublish;
-  if (prepublishHook) {
-    for (const pkg of packagesToPublish) {
-      const packageInfo = bumpInfo.packageInfos[pkg];
+
+  // finally pass through doing the actual npm publish command
+  for (const pkg of packagesToPublish) {
```
ecraig12345, line 48: I'm not sure if the change to order/grouping of operations here is right (but I can kind of see arguments for either ordering). @kenotron or @xugao what do you think?

Before:
prepublish all packages, publish all packages, (new) postpublish all packages

After:
for each package, prepublish/publish/postpublish

kenotron:
There should be some description of the use case for this feature, and also the approach in theory. For example, the comment was called out in the PR that there's now a new kind of grouping. In general, the prepub & maybe the postpub hook would be executed in phases. If a package level prepub and postpub steps are needed, I think npm lifecycle scripts are what is needed.

kenotron (changes requested):
Please add a description of what the use case is. Preferably in an issue as a feature request?

mmaclach:
Hi @kenotron 

I've added a feature request here: https://github.com/microsoft/beachball/issues/488

For the approach, I followed the example of the existing one and moved the hooks inside the main publish loop. Re: your comment of keeping the old grouping, I will go ahead and revert this grouping change.

src/publish/publishToRegistry.ts
mmaclach: Per the comment below I will go ahead and revert the grouping change.

src/types/BeachballOptions.ts
```diff
@@ -73,6 +73,12 @@ export interface RepoOptions {
      * repository.
      */
     prepublish?: (packagePath: string, name: string, version: string) => void | Promise<void>;
+
+    /**
+     * Runs for each package after the publish command.
+     *
```
ecraig12345, line 79:
```suggestion
```

ecraig12345 (approved):
Thanks for the updates!

ecraig12345:
@mmaclach In the future it's preferable to include a description in the PR description field itself, and just link to the existing issue (no need to create a new one). I updated it this time and cleared the request changes status.

ecraig12345:
> If a package level prepub and postpub steps are needed, I think npm lifecycle scripts are what is needed.

@kenotron I think the current approach is reasonable since it's already what we do for prepublish (you added it in #319).

src/types/BeachballOptions.ts
```diff
@@ -73,6 +73,11 @@ export interface RepoOptions {
      * repository.
      */
     prepublish?: (packagePath: string, name: string, version: string) => void | Promise<void>;
+
+    /**
+     * Runs for each package after the publish command.
```
ecraig12345, line 78:
```suggestion
     * Runs for each package after the publish command.
     * Any file changes made in this step will **not** be committed automatically.
```

mmaclach:
> @mmaclach In the future it's preferable to include a description in the PR description field itself, and just link to the existing issue (no need to create a new one). I updated it this time and cleared the request changes status.

sounds good, thanks @ecraig12345

kenotron:
this looks good to me as well

kenotron (approved):

Issue #490, opened 2021-02-09
feat: add more user friendly remote resolution for `change/check` command


Hotell:
Currently if `beachball check` is invoked it will run against `origin` remote by default. 

While this can be changed by explicitly specifying `--branch upstream/master` it would be worth it to add smarter remote resolution that is common to github contribution flow:
- is there a `upstream` remote ?
  - notify user that it's gonna run against `upstream` and explicitly tell him that to change this he needs to use `--branch` flag
- if yes use that by default
- otherwise go with `origin`

ecraig12345:
This is finally done (#34).

PR #504, opened 2021-03-04
Fix detection of default remote branch when it's not specified at CLI

ecraig12345:
#443 introduced an issue where beachball doesn't attempt to determine the default remote except if a branch is specified at the command line. Fix by adding branch defaulting logic to `getRepoOptions` (renamed from `getRootOptions` to match the interface name), since this is the easiest way to get the remote/branch mixed in with the correct order of precedence.

Partially addresses #490.


src/options/getPackageOptions.ts
```diff
@@ -12,7 +12,7 @@ import path from 'path';
 export function getCombinedPackageOptions(actualPackageOptions: Partial<PackageOptions>): PackageOptions {
   const defaultOptions = getDefaultOptions();
   const cliOptions = getCliOptions(process.argv);
-  const rootOptions = getRootOptions(cliOptions);
+  const rootOptions = getRepoOptions(cliOptions);
```
kubkon, line 15: This is entirely a stylistic nit, and as such, feel free to ignore. Here goes. To be extra consistent, do you think it would make sense to also rename this local binding to `repoOptions`? The two immediate benefits I can see are 1) consistency of the rename, and 2) less confusion for other potential contributors when browsing the sources.

```suggestion
  const repoOptions = getRepoOptions(cliOptions);
```

src/options/getPackageOptions.ts
```diff
@@ -12,7 +12,7 @@ import path from 'path';
...
   return {
     ...defaultOptions,
     ...rootOptions,
```
kubkon, line 18:
```suggestion
    ...repoOptions,
```

src/options/getRepoOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+  repoOptions = repoOptions || ({} as RepoOptions);
+
+  if (!cliOptions.branch) {
+    // Add or fix up the branch in repoOptions if it's not specified in cliOptions (which takes precedence)
+    if (repoOptions.branch && !repoOptions.branch.includes('/')) {
+      repoOptions.branch = getDefaultRemoteBranch(repoOptions.branch, cliOptions.path);
+    } else if (!repoOptions.branch) {
+      repoOptions.branch = getDefaultRemoteBranch('master', cliOptions.path);
+    }
+  }
```
kubkon, line 25: This is the bit that was added compared with the old implementation in `getRootOptions.ts`, am I correct? Could you explain why we check for existence of '/' in the branch name?

kubkon (approved):
LGTM, just a couple of questions/nit but they are not vital to landing this PR, so feel free to ignore.

src/options/getRepoOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  return repoOptions;
+}
+
+function tryLoadConfig(configPath: string): RepoOptions | null {
+  const configExplorer = cosmiconfigSync('beachball');
```
Hotell, line 31: **thought:** this will throw Error if no config is found. are we sure we don't wanna handle it ?

src/options/getRepoOptions.ts
```diff
@@ -0,0 +1,40 @@
...
+  const loadResults = configExplorer.load(configPath);
+  return (loadResults && loadResults.config) || null;
```
Hotell, line 33: **suggestion:** can we use ternary operator with explicit runtime checks? this is hard to read

src/options/getRepoOptions.ts
```diff
@@ -0,0 +1,40 @@
+import { cosmiconfigSync } from 'cosmiconfig';
+import { getDefaultRemoteBranch } from 'workspace-tools';
+import { RepoOptions, CliOptions } from '../types/BeachballOptions';
+
+export function getRepoOptions(cliOptions: CliOptions): RepoOptions {
+  let repoOptions: RepoOptions | null;
```
Hotell, line 6: **suggestion:** nesting ifs adds cognitive load and is hard to understand the program flow in general (also I don't see tests for this)

I'd recommend exit early approach with immutable approach.

Simple example:
> any kind of mutation adds complexity so this example is not ideal - primarily applied exit early strategy and removed nesting

```ts
function getRepoOptions(cliOptions: CliOptions){
  let repoOptions = tryLoadConfig(cliOptions.configPath);

  if(repoOptions === null){
    console.error(`Config file "${cliOptions.configPath}" could not be loaded`);
    process.exit(1);
  }

  repoOptions = trySearchConfig() || {} 

  if (!cliOptions.branch)
    const branch = resolveBranch(repoOptions,cliOptions)
    repoOptions.branch = branch
  }

  return repoOptions
}
```

src/options/getRepoOptions.ts
```diff
@@ -0,0 +1,40 @@
+import { cosmiconfigSync } from 'cosmiconfig';
+import { getDefaultRemoteBranch } from 'workspace-tools';
+import { RepoOptions, CliOptions } from '../types/BeachballOptions';
+
+export function getRepoOptions(cliOptions: CliOptions): RepoOptions {
```
Hotell, line 5: **suggestion** this is not a very good API, as we are lying to consumers, instead we should return proper interface or null

src/options/getPackageOptions.ts
ecraig12345: Good point, I just missed that since I did the function name change from another file.

src/options/getRepoOptions.ts
ecraig12345: This isn't a new file, just renamed, so I'm not necessarily looking to fix up things beyond the remote selection. For this specific issue, I don't see anything in the cosmiconfig docs about throwing if not found, and testing it out locally it doesn't appear to throw. https://www.npmjs.com/package/cosmiconfig

src/options/getRepoOptions.ts
ecraig12345: Agreed about the nesting in general but I'm not seeing a reasonable way to further simplify it while still handling all required cases.

src/options/getRepoOptions.ts
ecraig12345: As far as I can tell, the return type of this function actually ought to be `Partial<RepoOptions>` since specifying all the options in the config is never required. (Also all of the beachball option types are not the most crisp/accurate with their notion of required vs. optional properties.) But fixing that would potentially farther-reaching consequences than I want to deal with right now, since this isn't new code.

kenotron:
I'm approving this based on the fact that a refactor of the config options part can be refactored later. This is just moving code around.

kenotron (approved):

PR #511, opened 2021-04-12
Beachball change command - to commit only changefiles 

nickymunga:
1. This change bumps up [workspace-tools package](https://github.com/microsoft/workspace-tools) to v0.13.0. The commit utility in [workspace-tools package](https://github.com/microsoft/workspace-tools) now accepts [additional options](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---only) for the commit command. 
2. This change aims at taking advantage of this; & include the change files in the commit function call `commit -m "change files" {package-root}\\change\\beachball-05571989-7c13-4b9e-b9bf-60c0dea76b11.json`. This will ensure only the specified changefiles(s) are committed while other staged files will be ignored.

### DIFF
--- a/src/changefile/writeChangeFiles.ts
+++ b/src/changefile/writeChangeFiles.ts
@@ -41,8 +41,8 @@ export function writeChangeFiles(
     stage(changeFiles, cwd);
     if (commitChangeFiles) {
       // only commit change files, ignore other staged files/changes
-      const commitOptions = `${changeFiles.join(' ')}`;
-      commit('Change files', cwd, [commitOptions]);
+      const commitOptions = ['--only', path.join(changePath, '*.json')];
+      commit('Change files', cwd, commitOptions);
     }
 
     console.log(
--- a/src/changefile/writeChangeFiles.ts
+++ b/src/changefile/writeChangeFiles.ts
@@ -40,7 +40,9 @@ export function writeChangeFiles(
 
     stage(changeFiles, cwd);
     if (commitChangeFiles) {
-      commit('Change files', cwd);
+      // only commit change files, ignore other staged files/changes
+      const commitOptions = `${changeFiles.join(' ')}`;
+      commit('Change files', cwd, [commitOptions]);
     }
 
     console.log(
--- a/CHANGELOG.json
+++ b/CHANGELOG.json
@@ -2,6 +2,51 @@
   "name": "beachball",
   "entries": [
     {
+      "date": "Wed, 14 Apr 2021 15:34:23 GMT",
+      "tag": "beachball_v2.2.0",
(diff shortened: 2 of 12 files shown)


src/changefile/writeChangeFiles.ts
```diff
@@ -40,7 +40,9 @@ export function writeChangeFiles(
 
     stage(changeFiles, cwd);
     if (commitChangeFiles) {
-      commit('Change files', cwd);
+      // only commit change files, ignore other staged files/changes
+      const commitOptions = `${changeFiles.join(' ')}`;
```
kenotron, line 44: There are several things wrong with this:

1. The whole point of the `commit(msg, cwd, options)`'s new `options` param is to pass an array to the command. It should be just `const commitOptions = changeFiles;` according to the work done in `workspace-tools`.
2. Windows has an annoying problem of max length for a command: https://devblogs.microsoft.com/oldnewthing/20031210-00/?p=41553 - so it's a "bad idea" to shove a potentially growing list of changefiles through the command line like that.
3. You already staged the files with very specific files above. Essentially you're already doing what you intended to do. So I'm not sure if this change even solves the problems you're trying to solve. Have you tested your change before this PR?

Look at line 41-44:

```js
stage(changeFiles, cwd); // this line already staged files specifically for this commit: git add file1; git add file2; git add file3
if (commitChangeFiles) {
  commit("change files", cwd); // this line just calls git commit
}
```

In reading this your PR & your description and the workspace-tools change, I don't think you are solving what you intended.

kenotron (changes requested):
I am not sure what problem you're running into with this change? It seems like we already a bunch of work to stage the specific change files. Unless you're actually say that you stage other things during the PR, then you should consider doing `commit("change files", cwd, "change/*.json")` so you don't blow up the Windows max length on commands. (this is why `stage()` doesn't do it all in one command)

src/changefile/writeChangeFiles.ts
kenotron: actually I suggest we maybe should do `const commitOptions = ['--only', '${workspaceRoot}/change/*.json'];`

p01: Good idea about `const commitOptions = ['--only', '${workspaceRoot}/change/*.json'];`

The goal of this PR is to make `yarn change` only commit the `change/*.json` files even if other files are staged.

nickymunga:
@kenotron I made the suggested changes. Please review when you can.

p01:
Using `--only` is indeed cleaner. Thanks for the change @nickymunga  👍

kenotron (approved):
Ooops wrong option in the radio :)

PR #513, opened 2021-04-13
feat: Support generating change files from conventional commits

asmundg:
This adds optional support for parsing changefile properties from structured conventional commit messages. When --useConventionalCommits is set, and at least one commit message from the current branch adheres to the structured commits standard, the last conforming commit message is used to populate the change files' type and description.


kenotron:
This is a nice thought - one that could potentially be expounded upon. Say, the flow could even be "at the time of a git commit, if it has a certain format, generate the change file along with the commit." This way, you're even skipping the "beachball change" step!

kenotron:
more scenarios to consider:

1. one might want to configure additional types of conventions in the commits - so they would need a way to map the conventions to the semver changes
2. after this is in, another neat thing to do is to then allow for a changelog generator that follows conventional-changelog: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog/README.md
3. there is yet another neat tool called: https://github.com/tunnckoCoreLabs/parse-commit-message that can give you a headstart on being able to parse commit messages

Aaron-Pool:
@asmundg @kenotron Did this get any traction? I'm trying to find a good automated release/version handler as I migrate a library to monorepo structure, but I'm also a pretty big fan of conventional commits. So I'd prefer not to give that up, if possible.

asmundg:
@Aaron-Pool landing this on my backlog, but has kept getting pushed down. I am hoping to find some time this month.

Aaron-Pool:
@asmundg How much work is still outstanding? Is it in a state that I could play around with it? Or is it not usable at all?

asmundg:
@Aaron-Pool I use it daily. :) The outstanding work is mostly around cleaning up and looking at integrating into git hooks. As such, this change is self-contained, but we wanted to think some more about usage patterns before we just added it.

Aaron-Pool:
@asmundg, cool, I think I'll install your PR and play around with it. Also, small tidbit of initial feedback--shouldn't `chore` map to `None`, rather than `patch`? I was under the impression that `chore` mapped to things like documentation and file reorganization, things that shouldn't actually produce a new version?

src/changefile/conventionalCommits.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function map(d: ConventionalCommit): Change | undefined {
+  if (d.breaking) {
+    return { type: 'major', message: d.message };
+  }
+
+  switch (d.type) {
+    case 'fix':
+    case 'chore':
```
tido64, line 38: Why would `chore` map to `patch` here? When I think of `chore` commits, I usually think of things like bumping a dev dependency or other changes that does not affect the published package.

Also, can you add support `docs` and `test` as well?

Aaron-Pool: @tido64 I've been doing some hacking locally to get it "just right" and I think what would be best would be to let the user provide a dictionary of allowed types and the change type (major/minor/patch/none) it should map to. That has worked well for me.

tido64: Yes, that makes sense. Also, I just saw your comment now, asking the exact same question 🤣

src/changefile/conventionalCommits.ts
```diff
@@ -0,0 +1,44 @@
...
+  const match = commitMessage.match(COMMIT_RE);
+  const data: ConventionalCommit | undefined = match
+    ? { type: match[1], scope: match[2], breaking: !!match[3], message: match[4] }
+    : undefined;
+  return data && map(data);
+}
+
+function map(d: ConventionalCommit): Change | undefined {
+  if (d.breaking) {
+    return { type: 'major', message: d.message };
```
NickGerleman, line 33: There's a special case here, for semver 0.x.y packages. In that scheme breaking changes are a minor bump, and fixes/feats are a patch bump.

npm, dependabot, are aware of that scheme, it it's used in some common packages like react-native.

src/changefile/conventionalCommits.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  switch (d.type) {
+    case 'chore':
+      return { type: 'none', message: d.message };
+    case 'fix':
+      return { type: 'patch', message: d.message };
+    case 'feat':
+      return { type: 'minor', message: d.message };
```
NickGerleman, line 42: How does this intersect with prerelease packages?

src/changefile/conventionalCommits.ts
```diff
@@ -0,0 +1,44 @@
...
+  message: string;
+}
+
+interface Change {
+  type: ChangeType;
+  message: string;
+}
+
+export function parseConventionalCommit(commitMessage: string): Change | undefined {
+  const match = commitMessage.match(COMMIT_RE);
```
NickGerleman, line 24: Nit: consider doing case insensitive match

PR #520, opened 2021-05-12
Validate the dependentChangeType field

nickymunga:
Validate the dependentChangeType field in the changefile.
Validations include: 

- Check if it is a valid change type.
- Check if it is not a disallowed change type (if specified)

### DIFF
--- a/src/validation/validate.ts
+++ b/src/validation/validate.ts
@@ -96,7 +96,7 @@ export function validate(options: BeachballOptions, validateOptionsOverride?: Pa
       );
       process.exit(1);
     }
-    if (!change.dependentChangeType && !isValidDependentChangeType(change.dependentChangeType, disallowedChangeTypes)) {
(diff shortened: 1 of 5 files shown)


src/validation/validate.ts
```diff
@@ -95,6 +96,16 @@ export function validate(options: BeachballOptions, validateOptionsOverride?: Pa
       );
       process.exit(1);
     }
+    if (
+      !change.dependentChangeType ||
+      !isValidDependentChangeType(change.dependentChangeType) ||
+      disallowedChangeTypes?.includes(change.dependentChangeType)
```
kenotron, line 102: disallowedChangeTypes check should be passed in the "isValidDependencyChangeType" so it can do the filtering there

kenotron (changes requested):
Minor change in the implementation

kenotron (approved):

PR #521, opened 2021-05-20
npm: Add basic auth (_password) via new authType argument

JasonGore:
In order to use a PAT with beachball it needs to support basic authentication with npm. The pre-existing `_authToken` method is bearer auth and does not work with PATs.

This PR adds a new `authType` argument which defaults to existing behavior. However, when `password` is specified beachball now changes the npm auth argument to use basic authentication.


src/types/Auth.ts
```diff
@@ -0,0 +1 @@
+export type AuthType = 'authtoken' | 'password';
```
christiango, line 1: Just to double check , these terms are standard vocabulary in NPM? Or is this an Azure DevOs concept?

JasonGore: This is npm-specific vocabulary (`_authToken` vs `_password`) as `token` is only used for npm commands. I'm open to more "standardized" verbiage but couldn't think of anything that wouldn't also obfuscate the npm meaning.

src/__e2e__/publishE2E.test.ts
```diff
@@ -54,6 +54,7 @@ describe('publish command (e2e)', () => {
```
christiango, line 54: Any test coverage worth adding for the _password flow?

christiango (approved):

src/types/Auth.ts
christiango: Sounds good! Just was curious where these terms came from

src/__e2e__/publishE2E.test.ts
JasonGore: I had thought about it I don't see any pre-existing token e2e tests. The logic I added is actually pretty straightforward, but if Ken feels strongly about it we probably need to figure out how to add baseline tests as well.

kenotron: Nah, this is fine. If you want to add a test - it would only be the getNpmAuthArgs function, but that's like 90% just a code move. So I don't mind it not having tests (yet?)

kenotron (approved):

PR #530, opened 2021-05-31
chore: Fix a typo in the status message

evancharlton:
The status message says:
> git commited these change files: <list>

In fact, it should say "committed".


ecraig12345 (approved):
Thanks!

PR #533, opened 2021-06-14
Add star command and enable syncing from changelogs

yingzhe:
- Adds a `star` command that allows referencing workspace packages as a `*` dependency. This is not applicable to yarn 2 where `workspace:` notation exists.
- Adds ability for `sync` command to pull versions from changelogs


ecraig12345:
This PR is old and conflicted enough now that it would require a good bit of work to get back to a mergeable state, plus I'm not sure if this feature is appropriate for beachball to handle. If you're still interested in this, please open an issue to discuss the design.

PR #543, opened 2021-06-30
Fix for handling of custom schema while updating the CHANGELOG.json file

### LINKED ISSUE #549: CHANGELOG.json file does not contains all the custom fields present in the change file
Currently, the consumers have the option to specify a custom schema in the beachball config file `(changeFilePrompt: { changePrompt: prompt => {} })` which gets used for creating the individual change files. 

But when the CHANGELOG.json files are updated not all the fields of the change file gets included in the update written to the CHANGELOG.json file. Due to this the custom schema fields although being part of the generated change files are not included in the CHANGELOG.json file.

chandpravandan:
Currently, the consumers can specify a custom schema in the beachball config file (`
changeFilePrompt: {
    changePrompt: prompt => {}
}
`) which gets used for creating the individual change files. But when the CHANGELOG.json files are updated not all the fields of the change file gets included in the update written to the CHANGELOG.json file. Due to this the custom schema fields although being part of the generated change files are not included in the CHANGELOG.json file.

This PR introduces the fix to include all the custom schema fields present in the change file to be written to the CHANGELOG.json file when the CHANGELOG.json file is updated.

Fixes #549

### DIFF
--- a/change/beachball-7c2b234c-3dbf-430f-956b-a4bdb4fdacfa.json
+++ b/change/beachball-7c2b234c-3dbf-430f-956b-a4bdb4fdacfa.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Fix for handling of custom schema while updating the CHANGELOG.json file",
+  "packageName": "beachball",
+  "email": "pravcha@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


chandpravandan:
This PR is for resolving https://github.com/microsoft/beachball/issues/549

ecraig12345:
Looking at the implementation of `promptForChange`, I initially thought it was an accident that extra answers returned by the prompt were being included in the change file, but it appears this was intentionally added in #298 (without accompanying types or tests). I just pushed a change adding types and simplifying the implementation to use more idiomatic typescript.

ecraig12345 (approved):

PR #544, opened 2021-06-30
Changes to have custom renderer at the package level

chandpravandan:
Currently the beachball allows to specify a custom renderer for the CHANGELOG.md file at the repository level. So, we cannot have a custom renderer at the package level.

This PR introduces more flexibility in terms of having a custom renderer for the CHANGELOG.md file, by allowing developers to specify a custom renderer at the package level.

For example - 

Currently in the beachball config file we can have - 

`
changelog: {
    customRenderers: {}
}
`

which will be changed to - 

`
changelog: {
    packageName1: {
        customRenderers: {}
    },
    packageName2: {
        customRenderers: {}
    }
}
`

### DIFF
--- a/src/changelog/writeChangelog.ts
+++ b/src/changelog/writeChangelog.ts
@@ -134,7 +134,7 @@ async function writeChangelogFiles(
       previousContent,
       newVersionChangelog,
       isGrouped,
-      changelogOptions: options.changelog || {},
+      changelogOptions: options.changelog?.[`${newVersionChangelog?.name}`] || {},
     });
 
     fs.writeFileSync(changelogFile, newChangelog);


ecraig12345 (changes requested):
It looks like with your current change, specifying a per-package renderer will be *required* (breaking existing usage). Can you please update the PR to support either repo-wide or per-package renderers?

chandpravandan:
Are you suggesting to club this PR's changes with https://github.com/microsoft/beachball/pull/545, since both of them are sort of related.

ecraig12345:
> Are you suggesting to club this PR's changes with #545, since both of them are sort of related.

No, I mean that right now, if someone has `changelog: { customRenderers: {} }`, I think it will stop working after your change. (I'm not sure why the typings didn't catch this.)

Thinking about it more, I'm not sure that this change is needed at all for what you want to achieve (if I'm understanding right). The package name should already be accessible via `PackageChangelogRenderInfo.newVersionChangelog.name`, and then you can conditionally modify rendering as needed.

ecraig12345:
Closing for housekeeping

PR #545, opened 2021-06-30
Option to specify Custom changelog at the package level

chandpravandan:
Currently the beachball allows to specify a custom schema for the changelog(s) at the repository level. So, we cannot have a custom schema at the package level.

This PR introduces more flexibility in terms of defining a custom schema for the changelog, by allowing developers to specify a custom schema at the package level.

For example - 

Currently in the beachball config file we can have - 

`
changeFilePrompt: {
    changePrompt: prompt => {}
}
`

will be now like - 

`
changeFilePrompt: {
    changePrompt: (prompt, packageName) => {
        // logic to return the schema based on the packageName
    }
}
`


ecraig12345:
Thanks for doing this! Can you please add a change file using `yarn change` so it will be published?

chandpravandan:
Sure, I have added the change file in this PR.

ecraig12345 (approved):

PR #555, opened 2021-07-16
Fixes changelog comments being overridden by the auto-generated dependent bump comments

### LINKED ISSUE #553: Change file comments for dependents ignored as of 2.6.0
After recently updating beachball to 2.6.2 we recently noticed we were getting lots of generic `Bump <package> to <version>` entries in said package's changelog file, which is tautological and not very useful.

I set up [a basic repro](https://github.com/JasonGore/beachball-comments-repro) running publish across each beachball ver and branching the result. I verified this behavior seems to be happening to packages that are dependents of other packages. Change file comments are ignored and overwritten by this generic entry.

In each of the package update commit links below, you can see how this issue surfaces with `package2`, which is a dependent of `package1`:

| beachball version commit | Issue occurs |
|------------------------------|--------------|
| [repro 2.6.2 commit](https://github.com/JasonGore/beachball-comments-repro/commit/e1bd96b47a5b5dd14594f19d5495913f0594b8f4#diff-b28b1ed9821f38c64524ea13e19264bb0772f8376afffd5d8a1f8277621833eb) | Y |
| [repro 2.6.0 commit](https://github.com/JasonGore/beachball-comments-repro/commit/700b81ef83cacc986cbb77803a20e0e61d2b1f55#diff-b28b1ed9821f38c64524ea13e19264bb0772f8376afffd5d8a1f8277621833eb) | Y |
| [repro 2.5.1 commit](https://github.com/JasonGore/beachball-comments-repro/commit/28a71e9ed254f0271381714d5976d54387e622ff#diff-b28b1ed9821f38c64524ea13e19264bb0772f8376afffd5d8a1f8277621833eb) | N |

With 2.5.1 you can see the `package2` change file comment comes through, indicating that this is a regression with 2.6.0 and #531, which is the only change related to 2.6.0.

kenotron:
Fixes #553 

1. getPackageChangelogs was taking in the incorrect changeinfo for its job - changed back to dependentChangeInfos + the changeFileChangeInfos
2. fixes the updateRelatedChangeTypes to create  baseChangeInfo based on changeFileChangeInfos.get(changeFile) rather than package name
3. added some test cases for the updated code

### DIFF
--- a/src/__tests__/changelog/getPackageChangelogs.test.ts
+++ b/src/__tests__/changelog/getPackageChangelogs.test.ts
@@ -1,0 +1,71 @@
+import { getPackageChangelogs } from '../../changelog/getPackageChangelogs';
+import { BumpInfo } from '../../types/BumpInfo';
+import { ChangeSet } from '../../types/ChangeInfo';
+import { PackageInfos } from '../../types/PackageInfo';
+
+describe('getPackageChangelogs', () => {
+  it('should have multiple comment entries when a package has a changefile AND was part of a dependent bump', () => {
+    const changeFileChangeInfos: ChangeSet = new Map([
+      [
+        'foo.json',
+        {
+          comment: 'comment for foo',
+          commit: 'deadbeef',
+          dependentChangeType: 'patch',
+          email: 'something@something.com',
+          packageName: 'foo',
+          type: 'patch',
+        },
+      ],
+      [
+        'bar.json',
+        {
+          comment: 'comment for foo',
+          commit: 'deadbeef',
+          dependentChangeType: 'patch',
+          email: 'something@something.com',
+          packageName: 'bar',
+          type: 'patch',
+        },
+      ],
+    ]);
+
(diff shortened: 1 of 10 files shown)


src/__tests__/changelog/getPackageChangelogs.test.ts
```diff
@@ -0,0 +1,71 @@
...
+          dependentChangeType: 'patch',
+          email: 'something@something.com',
+          packageName: 'foo',
+          type: 'patch',
+        },
+      ],
+      [
+        'bar.json',
+        {
+          comment: 'comment for foo',
```
JasonGore, line 23: Should be `comment for bar`? Just wanted to point this out in case it's causing any false positives. Seems there are more below too

kenotron: Arrrgh, typo! But the results are the same - test is verifying counts of things based on keys not value mostly.

Issue #420, opened 2020-10-15
No changes detected in GitHub actions


NickSevens:
I'm setting up beachball in a github action PR trigger. But, no changes are ever detected when pushing a commit to the PR.
If I run `beachball change` locally the changes _are_ detected as should.

Can anyone point me in the direction of what I'm doing wrong?

The github action file looks like this:

```
name: CR validation

on:
  pull_request:
    branches: [master]

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v2

      - name: Use Node.js
        uses: actions/setup-node@v1
        with:
          node-version: "12.x"

      - name: Install dependencies
        run: yarn

      - name: Check
        run: yarn checkchange
```

Thanks in advance

NickSevens:
PS: I've run the same PR validation in Azure DevOps and there the changes _are_ detected, so it seems to only be faulty in GitHub actions? :/

kenotron:
The thing to do is add 

```
- uses: actions/checkout@v2
  with: 
    fetch-depth: 0
```

Something changed about github actions to disable the ability for beachball to automatically get mergebase via the `...` mechanism for diffs.

kenotron:
I'm open to ideas but this is the best I can do FOR NOW. We need a better solution, but not sure what yet.

NickSevens:
> The thing to do is add
> 
> ```
> - uses: actions/checkout@v2
>   with: 
>     fetch-depth: 0
> ```
> 
> Something changed about github actions to disable the ability for beachball to automatically get mergebase via the `...` mechanism for diffs.

This worked, thanks @kenotron

PR #558, opened 2021-07-20
Fix checkchange in PR workflow

ecraig12345:
Applying fix described in https://github.com/microsoft/beachball/issues/420#issuecomment-712465458 to beachball itself


ecraig12345:
PR build failed = the change worked 😝

PR #560, opened 2021-07-20
Update prettier

ecraig12345:
Update prettier and run on all files. I added the `arrowParens: 'avoid'` option to reduce formatting changes. 

Also changed the `prettier` command to be called `format` so that it's possible to run prettier itself with custom args.


package.json
```diff
@@ -21,7 +21,7 @@
     "checkchange": "node ./lib/cli.js check",
     "docs": "vuepress dev docs --host localhost",
     "docs:build": "vuepress build docs",
-    "prettier": "prettier --write --config ../../.prettierrc --ignore-path ../../.prettierignore 'src/**/*'",
+    "format": "prettier --write '**/*'",
```
ecraig12345, line 24: Having this command called `prettier` prevented running prettier directly with custom args. Switching to `format` makes that possible.

Issue #21, opened 2019-07-16
Pre-release should be a change type


acoates-ms:
If the package isn't already a prerelease package, cli should query what the prerelease tag should be.

Otherwise, it should just reuse the existing prerelease tag.

kenotron:
Thanks for filing this!

acoates-ms:
Note prerelease changetype has been added, but it only shows up on a package that is already a prerelease version. -- There is no other way to specify the prerelease tag.

PR #571, opened 2021-08-02
Always available prerelease tags

okjodom:
Addresses Issue #21 . The goal of this PR is to improve the UX for selecting and using prerelease change types.

When creating change files, allow selection of prerelease change types even in packages that have never been marked prerelease. Should the prerelease change type be selected, cli will have a further prompt for prerelease prefix to be used. This prerelease prefix has the following behaviors:
 - It is recorded in a change file alongside change types.
 - It takes precedence over any previously used prefix.
 - It takes precedence prefix declared as part of options.

Experience changes:
1. Prerelease change type is available for selection:
image

2. If prerelease change type is selected, prompt for a prerelease tag
image

3. Write prerelease change type and tag to change file
image
image


src/changefile/promptForChange.ts
```diff
@@ -29,20 +28,19 @@ export async function promptForChange(options: BeachballOptions) {
 
     const disallowedChangeTypes = getDisallowedChangeTypes(pkg, packageInfos, packageGroups);
     const packageInfo = packageInfos[pkg];
-    const showPrereleaseOption = prerelease(packageInfo.version);
     const changeTypePrompt: prompts.PromptObject<string> = {
       type: 'select',
       name: 'type',
       message: 'Change type',
       choices: [
-        ...(showPrereleaseOption ? [{ value: 'prerelease', title: ' [1mPrerelease[22m - bump prerelease version' }] : []),
```
ecraig12345, line 38: I'm not sure that enabling prerelease by default will be desirable for everyone. One possible workaround is to add a separate option for always allowing prerelease, though that increases option bloat and is less discoverable for people who do want the feature. @kenotron any thoughts?

ecraig12345 (changes requested):
Following up on old PRs--I can see how having the option to have beachball introduce a prerelease version could make sense, but it's definitely not going to be desirable for all current beachball users to suddenly have that option exposed in all cases.

I can see a couple possibilities for how to address this:
- Update `disallowedChangeTypes` in `getDefaultOptions` to be `['prerelease']`. However this gets messy because we probably don't want to block people from bumping their existing prerelease versions if they don't have `disallowedChangeTypes` overridden.
- Add an option `alwaysAllowPrerelease` or something like that. Downside is this adds option bloat and feels redundant with `disallowedChangeTypes` (except slightly different).

ecraig12345:
Closing this since the PR is old, has conflicts, and needs changes. If you (or anyone else) would like to come back to this change, please comment on #21 so we can figure out the "right" design first.

Issue #579, opened 2021-08-12
Renaming change files and running ` beachball bump` with the --keep-change-files flag bumps up the package version


LarryKiniu:
Renaming change files and running beachball bump command with the --keep-change-files flag bumps up the package version.
Manually reverting PRs, where git deletes change files, and recreating them using `beachball change`  makes git mark these as renamed and might result in package versions changing in an unpredictable manner.

PR #580, opened 2021-08-12
Check renamed change files

LarryKiniu:
This PR checks whether change files have renamed and might solve #579

### DIFF
--- a/src/validation/areChangeFilesDeleted.ts
+++ b/src/validation/areChangeFilesDeleted.ts
@@ -37,7 +37,7 @@ export function areChangeFilesDeleted(options: BeachballOptions): boolean {
 
   if (changeFilesDeleted) {
     const changeFiles = changeFilesDeletedSinceRef.map(file => `- ${file}`);
-    const errorMessage = 'The following change files were deleted:';
+    const errorMessage = 'The following change files were deleted or renamed:';
 
     console.error(`${errorMessage}\n${changeFiles.join('\n')}\n`);
   }
--- a/src/validation/areChangeFilesDeleted.ts
+++ b/src/validation/areChangeFilesDeleted.ts
@@ -17,12 +17,12 @@ export function areChangeFilesDeleted(options: BeachballOptions): boolean {
     process.exit(1);
   }
 
-  console.log(`Checking for deleted change files against "${branch}"`);
+  console.log(`Checking for deleted and renamed change files against "${branch}"`);
   const changeFilesDeletedSinceRef = getChangesBetweenRefs(
     branch,
     'HEAD',
     [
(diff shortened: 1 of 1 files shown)


src/validation/areChangeFilesDeleted.ts
```diff
@@ -17,12 +17,12 @@ export function areChangeFilesDeleted(options: BeachballOptions): boolean {
...
   }
 
-  console.log(`Checking for deleted change files against "${branch}"`);
+  console.log(`Checking for deleted and renamed change files against "${branch}"`);
   const changeFilesDeletedSinceRef = getChangesBetweenRefs(
     branch,
     'HEAD',
     [
-      '--diff-filter=D', // showing only deleted files from the diff.
+      '--diff-filter=DR', // showing only deleted and renamed change files from the diff.
```
VincentBailly, line 25: Can we add some unit tests?

LarryKiniu: Sure - I will work on this.

arazabishov: +1 for tests. You can find existing tests for this function here: https://github.com/microsoft/beachball/blob/master/src/__e2e__/validation.test.ts#L86

LarryKiniu: Added some unit tests as well.

src/validation/areChangeFilesDeleted.ts
```diff
@@ -17,12 +17,12 @@ export function areChangeFilesDeleted(options: BeachballOptions): boolean {
     process.exit(1);
   }
 
-  console.log(`Checking for deleted change files against "${branch}"`);
+  console.log(`Checking for deleted and renamed change files against "${branch}"`);
```
arazabishov, line 20: I think it would make sense to rename the function too: `areChangeFilesDeletedOrRenamed`. A bit on verbose side, but reflects the purpose of the function :)

LarryKiniu: I suppose that would be more descriptive. However, it will break the API. There are a couple of options I can see to avoid this:
1. Create a new more descriptive function while keeping the current one and marking it for deprecation in the future.
2. Create a new method that just checks for renamed change files and calling it based on a flag like `--disallow-renamed-change-files` which can be used in conjunction with the current function.
Which option should we go with - or are there other options?

arazabishov: I think option #2 makes more sense :)

ecraig12345: @LarryKiniu This is an internal API, so you don't need to worry about breaking it.

PR #584, opened 2021-08-27
support multiple packages per changefile

jcreamer898:
In large repos when you're changing a bunch of packages at once, it can be kind of expensive to dump a bunch of JSON files into a single folder.

It's easy enough to just group the change files into single files expressing multiple packages per file.

Essentially, there would be a new `multiPackageChanges` option that would generate change files like...

```json
{
"changes": [{
  "packageName": "beachball",
  "type": "minor",
  "dependentChangeType": "patch",
  "email": "jcreamer@microsoft.com",
  "comment": "support multiple changes per changefile"
  }]
}
```

And there's a bit of confusion on my end as to why each `ChangeInfo` needs its own `comment` as I think that `comment` field is generated from the `yarn change` command? So perhaps we can bubble that `comment` field up into the `summary` as seen above.

Other than supporting multiple packages, in theory nothing else here functionally changes, other than I added the `human-id` library to give the change files a shorter and cleaner name.

### DIFF
--- a/change/change-chatty-cloths-develop.json
+++ b/change/change-chatty-cloths-develop.json
@@ -1,0 +1,9 @@
+{
+"changes": [{
+  "packageName": "beachball",
+  "type": "minor",
+  "dependentChangeType": "patch",
+  "email": "jcreamer@microsoft.com",
+  "comment": "support multiple changes per changefile"
+  }]
+}
--- a/src/changefile/readChangeFiles.ts
+++ b/src/changefile/readChangeFiles.ts
@@ -1,4 +1,4 @@
-import { ChangeSet, ChangeInfo } from '../types/ChangeInfo';
+import { ChangeSet, ChangeInfo, ChangeInfoMultiple } from '../types/ChangeInfo';
 import { getChangePath } from '../paths';
 import fs from 'fs-extra';
 import path from 'path';
@@ -53,7 +53,8 @@ export function readChangeFiles(options: BeachballOptions, packageInfos: Package
 
   filteredChangeFiles.forEach(changeFile => {
     const changeFilePath = path.join(changePath, changeFile);
-    let changeInfo: ChangeInfo;
+
+    let changeInfo: ChangeInfo | ChangeInfoMultiple;
     try {
       changeInfo = fs.readJSONSync(changeFilePath);
     } catch (e) {
@@ -70,10 +71,16 @@ export function readChangeFiles(options: BeachballOptions, packageInfos: Package
         return;
       }
     }
+    const changes: ChangeInfo[] = changeInfo.changes || [changeInfo as ChangeInfo];
 
-    if (scopedPackages.includes(changeInfo.packageName)) {
-      changeSet.set(changeFile, changeInfo);
+    for (const change of changes) {
+      const packageName = (change as ChangeInfo).packageName;
+
+      if (scopedPackages.includes(packageName)) {
+        changeSet.set(changeFile, change as ChangeInfo);
+      }
     }
   });
+
   return changeSet;
 }
--- a/src/changefile/writeChangeFiles.ts
+++ b/src/changefile/writeChangeFiles.ts
@@ -14,7 +14,8 @@ export function writeChangeFiles(
     [pkgname: string]: ChangeFileInfo;
   },
   cwd: string,
-  commitChangeFiles = true
+  commitChangeFiles = true,
+  groupChanges = false
 ): string[] {
   if (Object.keys(changes).length === 0) {
     return [];
@@ -24,6 +25,34 @@ export function writeChangeFiles(
   const branchName = getBranchName(cwd);
   if (changePath && !fs.existsSync(changePath)) {
     fs.mkdirpSync(changePath);
+  }
+
+  const prefix = 'change';
+  if (groupChanges) {
+    if (changes && branchName && changePath) {
+      const changeArray = Object.keys(changes).map(change => {
+        return changes[change];
+      });
+      const fileName = `${prefix}-${uuidv4()}.json`;
+      let changeFile = path.join(changePath, fileName);
+      const changeFiles = [changeFile];
+
+      fs.writeFileSync(changeFile, JSON.stringify({ changes: changeArray }, null, 2));
+
+      stage([changeFile], cwd);
+      if (commitChangeFiles) {
+        // only commit change files, ignore other staged files/changes
+        const commitOptions = ['--only', path.join(changePath, '*.json')];
+        commit('Change files', cwd, commitOptions);
+      }
+
+      console.log(
+        `git ${commitChangeFiles ? 'committed' : 'staged'} these change files: ${changeFiles
+          .map(f => ` - ${f}`)
+          .join('\n')}`
+      );
+      return [changeFile];
+    }
   }
 
   if (changes && branchName && changePath) {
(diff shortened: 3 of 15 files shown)


VincentBailly:
Do the suggestion of using toml have to be coupled with the suggestion to use a single files?
Coupling the two may dilute the core proposal, which is to use only one change file per PR.

jcreamer898:
Good call, maybe I can accomplish the same thing, but with an array of JSON in one file.

jcreamer898:
Updated @VincentBailly

ecraig12345 (changes requested):
Thanks for the contribution! I have some questions though. Also I'd definitely like to see some tests covering the new feature.

> In large repos when you're changing a bunch of packages at once, it can be kind of expensive to dump a bunch of JSON files into a single folder.

Expensive in what sense? If you mean it makes beachball commands slow to run, I'm not sure that the number of change files is actually the root cause (there are a couple other possibilities I'm aware of).

> And there's a bit of confusion on my end as to why each ChangeInfo needs its own comment as I think that comment field is generated from the yarn change command? So perhaps we can bubble that comment field up into the summary as seen above.

It depends on the scenario--sometimes using the same message across multiple packages makes sense, but other times it's good to be more granular. By default beachball asks for an individual message for each package, and the `--message` option is available as an override to give a single message for all packages.

> Other than supporting multiple packages, in theory nothing else here functionally changes, other than I added the human-id library to give the change files a shorter and cleaner name.

Whether this is actually an improvement is debatable, so I'd prefer it be discussed/handled separately.

jcreamer898:
Thanks for the reply @ecraig12345!

I added a few e2e tests to the PR now.

Expensive in this case was for git. In 1JS in MS Office, we have a folder with lots and lots of change files (that's a different topic), in any case, I did some research that shows that having a lot of files in one folder is really bad for git history and repo size, perf, etc. https://gist.github.com/jcreamer898/54f91122a5b64907eb02da0dc3c78b26

Thanks for the clarity on the message thing. It didn't end up mattering for my change.

Ken and I discussed the `human-id` change, and he was on board with it.

Thanks again!

kenotron (dismissed):

jcreamer898:
Need new approval here @kenotron

src/changefile/writeChangeFiles.ts
```diff
@@ -3,7 +3,7 @@ import { getChangePath } from '../paths';
 import { getBranchName, stage, commit } from 'workspace-tools';
 import fs from 'fs-extra';
 import path from 'path';
-import { v4 as uuidv4 } from 'uuid';
+import humanId from 'human-id';
```
ecraig12345, line 6: Would it be possible to add a repo option `humanId?: boolean` and only use `human-id` if that's enabled? I don't think it's desirable to name change files with random words in all cases (I wouldn't want this), and regardless, suddenly changing to that might catch people off guard.

src/types/ChangeInfo.ts
```diff
@@ -20,6 +20,11 @@ export interface ChangeInfo extends ChangeFileInfo {
   commit: string;
 }
 
+export type ChangeInfoMultiple = {
```
ecraig12345, line 23: nit, can you use an interface here to match the convention in this repo?

VincentBailly (approved):

ecraig12345 (changes requested):
Looked at this again and I have some more comments.

Also does this cause any changes to the *default* way change files are written? I'm not seeing this option enabled for beachball itself (unless I'm missing something) but it appears to be used in this PR's change file.

src/types/BeachballOptions.ts
```diff
@@ -42,6 +42,7 @@ export interface CliOptions {
   prereleasePrefix?: string | null;
   configPath?: string;
   commit?: boolean;
+  multipleChangesPerChangefile?: boolean;
```
ecraig12345, line 45: This name is a bit verbose, especially if it has to be specified on the command line. I wonder if it would be possible to come up with something more concise but still clear? 

Whatever the name is, it should probably be added to `RepoOptions` in addition or instead (IMO this would make more sense to enable in a config file than on the command line), and documented in `docs/overview/configuration.md`.

kenotron: I agree with the comment here about putting in "RepoOptions" and generally be checked into a beachball.config.js rather than in a CLI...

src/changefile/writeChangeFiles.ts
jcreamer898: I got this idea from https://github.com/atlassian/changesets/blob/eae72367e9cb8e708a1d441770dbc70930d4e1ad/packages/write/src/index.ts#L17 and personally find it much more readable than a guid, but that's up to you and @kenotron.

src/types/BeachballOptions.ts
jcreamer898: Sounds great. Thanks.

src/changefile/readChangeFiles.ts
```diff
@@ -71,9 +72,20 @@ export function readChangeFiles(options: BeachballOptions, packageInfos: Package
       }
     }
 
-    if (scopedPackages.includes(changeInfo.packageName)) {
-      changeSet.set(changeFile, changeInfo);
+    if (changeInfo.changes && changeInfo.changes.length) {
```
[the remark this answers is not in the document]
jcreamer898: Oo, nice.

src/types/ChangeInfo.ts
jcreamer898: Good call, not sure what I was thinking. Ha.

src/changefile/writeChangeFiles.ts
kenotron: I think this is completely subjective. The ONLY reason why I wouldn't do it is the addition of yet another dependency. Maybe we should just revert it for now and add it as a neat nice-to-have feature?

jcreamer898: I'll create a separate PR to talk about that. Don't wanna block this one on this.

jcreamer898:
Updated the option to `groupChanges`, and added docs, updated tests.

src/changefile/writeChangeFiles.ts
ecraig12345: > I think this is completely subjective. The ONLY reason why I wouldn't do it is the addition of yet another dependency. Maybe we should just revert it for now and add it as a neat nice-to-have feature?

Definitely agree that it's subjective (and I'm not concerned about the extra dep). It just seems like an especially polarizing type of subjective, as demonstrated by both of you liking it and me thinking "eew this is weird."

change/change-chatty-cloths-develop.json
```diff
@@ -0,0 +1,9 @@
+{
+"changes": [{
```
ecraig12345, line 2: One more question (sorry)--does this PR change the default way change files are written? I didn't see the new option being enabled for beachball itself, but it looks like this change file is using the new format.

jcreamer898: The update supports the multi-change format in the readChangefile, but the default in writeChangefile will still be regular way it is now.

jcreamer898:
Also, don't apologize, I truly appreciate your thoroughness. :)

ecraig12345 (approved):

ecraig12345:
Thanks for all the work on this! I think I'm actually going to enable it in our repo as well because it will make PR reviews easier.

jcreamer898:
Nice! How do I actually merge the change now? I have a not authorized message. :)

ecraig12345:
Oh sorry, didn't realize you didn't have permission

PR #586, opened 2021-09-02
Option to specify custom transform method for the change files

chandpravandan:
This PR introduces an option to specify any custom transform method for the change files. It can be used to add more information to the change files which might not be present at the time of their generation. 

The consumer can have the following beachball config file to use this functionality - 

```
{
       ...

      transform: {
          changeFiles: (changeFile, path) => {
          // Add or Edit the information in the changeFile
          return changeFile;
          };
      }

       ...
}
```

### DIFF
--- a/src/__e2e__/changelog.test.ts
+++ b/src/__e2e__/changelog.test.ts
@@ -191,6 +191,7 @@ describe('changelog generation', () => {
     });
 
     it('Verify that the changeFile transform functions are run, if provided', async () => {
+      const editedComment: string = 'Edited comment for testing';
       const monoRepo = monoRepoFactory.cloneRepository();
       monoRepo.commitChange('foo');
       writeChangeFiles({ foo: getChange() }, monoRepo.rootPath);
@@ -201,8 +202,11 @@ describe('changelog generation', () => {
       const beachballOptions = {
         path: monoRepo.rootPath,
         transform: {
-          changeFiles: (changeFile, path) => {
-            changeFile.comment = 'Edited comment for testing';
+          changeFiles: (changeFile, changeFilePath) => {
+            // For test, we will be changing the comment based on the package name
+            if(changeFile.packageName === 'foo'){
+              changeFile.comment = editedComment;
+            }
             return changeFile;
           }
         },
@@ -220,17 +224,14 @@ describe('changelog generation', () => {
       const packageInfos = getPackageInfos(monoRepo.rootPath);
       const changes = readChangeFiles(beachballOptions, packageInfos);
 
-      await writeChangelog(beachballOptions, changes, {}, packageInfos);
-
-      // Validate changelog for bar package
-      const barChangelogFile = path.join(monoRepo.rootPath, 'packages', 'bar', 'CHANGELOG.md');
-      const barChangelogText = fs.readFileSync(barChangelogFile, { encoding: 'utf-8' });
-      expect(cleanMarkdownForSnapshot(barChangelogText)).toMatchSnapshot();
-
-      // Validate grouped changelog for foo and bar packages
-      const groupedChangelogFile = path.join(monoRepo.rootPath, 'packages', 'foo', 'CHANGELOG.md');
-      const groupedChangelogText = fs.readFileSync(groupedChangelogFile, { encoding: 'utf-8' });
-      expect(cleanMarkdownForSnapshot(groupedChangelogText)).toMatchSnapshot();
+      // Verify that the comment of only the intended change file is changed
+      for(const [changeFileName, changeInfo] of changes){
+        if(changeFileName.substr(0, 3) === 'foo'){
+          expect(changeInfo.comment).toBe(editedComment);
+        }else{
+          expect(changeInfo.comment).toBe('comment 2');
+        }
+      }
     });
   });
 });
--- a/src/__e2e__/changelog.test.ts
+++ b/src/__e2e__/changelog.test.ts
@@ -202,7 +202,7 @@ describe('changelog generation', () => {
         path: monoRepo.rootPath,
         transform: {
           changeFiles: (changeFile, path) => {
-            changeFile.testParam = 'testParam';
+            changeFile.comment = 'Edited comment for testing';
             return changeFile;
           }
         },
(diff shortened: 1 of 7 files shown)


chandpravandan:
This PR is for - https://github.com/microsoft/beachball/issues/587

JasonGore:
I think it looks good. Does it also get called if beachball fetches another round of new change files during publish?

@kenotron I recommended a change like this for beachball so that we could apply an optional transform to beachball change files without having to modify files on disk or risk missing out on new change files being merged during published (which would happen unless this feature was a part of beachball). Having a transform like this was an idea I had to get around all of these issues. What do think about merging a feature like this in?

src/__e2e__/changelog.test.ts
```diff
@@ -189,5 +189,48 @@ describe('changelog generation', () => {
...
+        changelog: {
+          groups: [
+            {
+              masterPackageName: 'foo',
+              changelogPath: path.join(monoRepo.rootPath, 'packages', 'foo'),
+              include: ['packages/foo', 'packages/bar'],
+            },
+          ],
+        },
+      } as BeachballOptions;
```
JasonGore, line 218: Is this to work around typing issues? Curious if `const beachballOptions: BeachballOptions` didn't work

src/__e2e__/changelog.test.ts
```diff
@@ -189,5 +189,48 @@ describe('changelog generation', () => {
...
+
+      monoRepo.commitChange('bar');
+      writeChangeFiles({ bar: getChange({ packageName: 'bar', comment: 'comment 2' }) }, monoRepo.rootPath);
+
+      const beachballOptions = {
+        path: monoRepo.rootPath,
+        transform: {
+          changeFiles: (changeFile, path) => {
+            changeFile.comment = 'Edited comment for testing';
+            return changeFile;
```
JasonGore, line 206: Maybe only modify one change file based on `path` and confirm only expected change files change. I assume other tests cover test cases where `transform` and/or `changeFiles` is not provided, is that true?

src/types/BeachballOptions.ts
```diff
@@ -83,6 +83,10 @@ export interface RepoOptions {
      */
     postpublish?: (packagePath: string, name: string, version: string) => void | Promise<void>;
   };
+
+  transform?: {
```
JasonGore, line 87: I think some documentation would be good here like `hooks` above

src/__e2e__/changelog.test.ts
chandpravandan: `const beachballOptions: BeachballOptions` does a more stricter type check and we then had to provide all the mandatory fields. So, chose doing this way.

src/__e2e__/changelog.test.ts
chandpravandan: I have modified the test case to checking that the intended changeFile is transformed. I suppose we have a `bumps only packages with change files` test in the bump.test.ts file for this.

src/types/BeachballOptions.ts
chandpravandan: Yes, sure. I have added a description like we have for the `hooks`.

chandpravandan:
> I think it looks good. Does it also get called if beachball fetches another round of new change files during publish?
> 
> @kenotron I recommended a change like this for beachball so that we could apply an optional transform to beachball change files without having to modify files on disk or risk missing out on new change files that merged and get pulled during publish (which would happen unless this feature was a part of beachball). Having a transform like this was an idea I had to get around all of these issues. What do think about merging a feature like this in?

I looked at the publish flow and I found that after creating the temporary publish branch, we call `gatherBumpInfo` https://github.com/microsoft/beachball/blob/6a417b3a6590855e2788a5c2084292106e34731e/src/commands/publish.ts#L59. In this function we read the changeFiles from the disk and the custom transform method would be called (in the `readChangeFiles` method). Also, the `bumpInfo` retrieved from this function is used in the `bumpAndPush` (in which we fetch the changes from master).

Also, I think `readChangeFiles` is the util which will be called anytime we want to parse the changeFiles. So the custom transform method will be called any time we want to get the data from the changeFiles and then use it during `bump`.

src/__e2e__/changelog.test.ts
JasonGore: In that case you should do `const beachballOptions: Partial<BeachballOptions>` as it is still more typesafe than casting. It will flag any incorrectly typed or extraneous property names.

JasonGore (approved):

kenotron:
We have a hooks config for things like this potentially - I like the idea of having a transform, but if you just declared a transform, it isn't very clear when in the process you would be running this. I would prefer simply add a onPreBump or something like that.

kenotron (changes requested):
Can we attack this by adding a  "onPreBump" instead? We can then re-use an existing construct.

chandpravandan:
> Can we attack this by adding a "onPreBump" instead? We can then re-use an existing construct.

In my opinion the config parameter - `transform` indicates that it is intended for modifying an entity in-memory and through the type of the transform parameter we are conveying to the user that we will now be using this updated returned value for our further processing.

I also think it is different from the `onPreBump` hook because the `onPreBump` hook has a notion that the user is provided with an opportunity to perform a task before the bump happens but does not knows exactly when that hook will be called and also kind of encourages the consumers to make these changes on the disk (which is relatively an expensive and not so fast process) in the hope that only then they will be incorporated in further processing. 

Also, if we compare it with the current hooks we have for publish they are more intended for providing the user an opportunity to execute some task before and after the publish happens but they are not intended for consuming the modifications done in them. If we have `onPreBump` hook on the similar lines and call them just before the bump happens then we will also not be able to incorporate the changes done in the changeFiles in the hook since that would be called after we have read the changeFiles.

So, I think since we are trying the support to transform the changeFiles in-memory so that the user has the capability to edit them therefore `transform` term fits better. `onPreBump` hook is intended for providing different kind of support and in future we can perhaps add that also in beachball.

JasonGore:
Unless `onPreBump` can have hooks that are called as items like change files are processed so that they can be transformed "in memory" (rather than requiring modifying things on disk) then I agree with @chandpravandan. (I'm also not sure if new change files being fetched during publish count as `onPreBump` either, but we definitely want these change files covered as well.)

kenotron:
Point taken, I think you're both right. I think I had some hesitance towards having more options added that are the beginnings of an extension point of beachball - as we're seeing more needs here for transforms and such. But this is a perfect beginning of transforms of many kinds (maybe even includes global level changelogs, etc.)

kenotron (approved):

PR #591, opened 2021-09-08
feat(mergeChangelogs): Don't include dependent change entries in groupped changelogs

ling1726:
In groupped changelog, the dependent changelog entries don't add too much value when you're reading a changelog. Now these dependent change entries are no longer included in a groupped changelog
Before
```
master package changelog:

- foo: some cool changes
- baz: some cool changes
- bar: bump baz to version x
- other: bump baz to version x
- other: bump baz to version x
- other: bump baz to version x
- other: bump baz to version x
```

After
```
master package changelog:

- foo: some cool changes
- baz: some cool changes
```

- New option `ignoreDependentChanges` to change log group options
- `mergeChangelogs` can now ignore dependent change (version bump) entries
- Adds a `dependentChange` boolean flag to `ChangeInfo` to determine a change entry is from a dependent change
- Sets `dependentChange` flag in `updateRelatedChangeType` when generating change entries for dependent changes


src/__e2e__/changelog.test.ts
```diff
@@ -153,6 +155,49 @@ describe('changelog generation', () => {
       expect(cleanMarkdownForSnapshot(groupedChangelogText)).toMatchSnapshot();
     });
 
+    it.only('generates grouped changelog without dependent change entries', async () => {
```
ecraig12345, line 158:
```suggestion
    it('generates grouped changelog without dependent change entries', async () => {
```

ecraig12345 (approved):

ecraig12345 (commented):
Actually one more thing, can you add a test for if a package has normal changes AND would be a dependent change?

ling1726:
> Actually one more thing, can you add a test for if a package has normal changes AND a dependent change?

sure

src/__e2e__/__snapshots__/changelog.test.ts.snap
```diff
@@ -212,3 +212,58 @@ This log was last generated on (date) and should not be manually modified.
...
+<!-- Start content -->
+
+## 1.3.4
+
+(date)
+
+### Patches
+
+- comment 1 (test@testtestme.com)
+- Bump baz to v1.3.5 (test@testtestme.com)
```
ecraig12345, line 230: Isn't this supposed to be omitted? Unless I'm misunderstanding the test setup.

ecraig12345: never mind

ecraig12345 (dismissed):

kenotron:
In grouped changelogs, is there ever a good reason to have those entries? If not, then let's just remove the logs __always__

src/changelog/getPackageChangelogs.ts
```diff
@@ -22,7 +22,7 @@ export function getPackageChangelogs(
   const commit = getCurrentHash(cwd) || 'not available';
 
   for (let change of changeInfos) {
```
kenotron, line 24: I think the only place that this dependentChange flag is used is here. We already know which changeInfo is dependentChange because of the concat above. We could just break apart the loop here into two (changeFileChangeInfos + dependentChangeInfos). Then only execute the 2nd loop if the new "ignore flag" is true.

Further, I'm not convinced I would ever include the dependentChangeInfo for grouped messages, so maybe we just don't even pass that in at all from the writeGroupedChangelog() in the first place?

kenotron (changes requested):

ling1726:
> In grouped changelogs, is there ever a good reason to have those entries? If not, then let's just remove the logs **always**

It is technically a breaking change to do that, are we ok with this ?

kenotron:
Totally fine to major bump a behavior change in my book. But you'll have to do a "minor" first and change the changefile

src/changelog/getPackageChangelogs.ts
ling1726: That's a good callout, it makes the whole thing a lot easier, and the tests still pass as expected 👍

JasonGore:
I originally created #553 which was partially fixed but we're still having issues with tautological entries overwriting changefile comments in certain cases. I can work on a repro again but I'm concerned the correct approach for this PR should actually be a bug fix, not a feature to avoid.

JasonGore:
I've created #596 to highlight the issues we've been seeing. @ling1726 If these comments were instead "Dependent X was bumped" (as it was in v2.5.1) would that be more useful to you and would you still want this feature disabling them?

src/changelog/mergeChangelogs.ts
```diff
@@ -26,8 +26,11 @@ export function mergeChangelogs(
 
   changelogs.forEach(changelog => {
     (Object.keys(changelog.comments) as ChangeType[]).forEach(changeType => {
+
+
       if (changelog.comments[changeType]) {
-        result.comments[changeType] = (result.comments[changeType] || []).concat(changelog.comments[changeType]!);
```
kenotron, line 30: Unless I'm reading this wrong, but there's no meaningful changes to this file anymore, maybe you can just revert this file change.

src/changelog/getPackageChangelogs.ts
kenotron: I'm not convinced we need changes to the "dependentChange" tracking at all anymore... so this isn't needed?

src/__tests__/bump/updateRelatedChangeType.test.ts
```diff
@@ -83,6 +83,9 @@ describe('updateRelatedChangeType', () => {
     expect(bumpInfo.calculatedChangeInfos['foo'].type).toBe('minor');
```
kenotron, line 83: I think these tests can be reverted...

src/bump/updateRelatedChangeType.ts
```diff
@@ -151,6 +151,7 @@ export function updateRelatedChangeType(
     }
```
kenotron, line 151: I don't think we need this tracking here anymore either...

src/types/ChangeLog.ts
```diff
@@ -3,9 +3,9 @@
  * Changes made to interfaces here can affect custom changelog rendering done by end-user.
```
kenotron, line 3: We shouldn't need this type change either

kenotron (changes requested):

src/changelog/getPackageChangelogs.ts
ling1726: Yup removed, failed to check in the commit 😅

kenotron (approved):
This PR is in great shape now

PR #593, opened 2021-09-09
Add ignorePatterns option

### LINKED ISSUE #563: Feature: ignore changes to certain files
When checking if a change file is required, it would be nice to have an option to ignore changes in files which won't be published to npm. ~~This might be too expensive to enable by default in a large monorepo, but for small repos (such as beachball itself) it could be fine.~~

~~Note that any implementation would need to check both `"files"` and `.npmignore`--there's probably an npm package for this.~~

Thinking about this more, it's safer to have a list of ignore patterns specifically for beachball rather than a blanket assumption that things included in npmignore don't affect the published package. Most obvious example is that `src` in a typescript project is not published but OBVIOUSLY changes the published result.

ecraig12345:
Add an `ignorePatterns` option which takes a list of minimatch patterns for files to ignore changes in when determining changed packages. This is most likely to be relevant in non-monorepo scenarios where config files live at the package root.

The immediate motivation is dependabot PRs such as #590 which only modify yarn.lock but are blocked due to missing change files (when a change file isn't even needed).

Fixes #563


kenotron (approved):
Excellent feature! I was hoping to make this when I had the time, but this is super useful!

PR #599, opened 2021-09-20
chore(BeachballOptions): Move `scope` to `RepoOptions` type

ling1726:
The `scope` CLI option, can be used in the configuration file regardless
of it being a cli specific option.

Adds `scope` to the `RepoOptions` type so that it can be used in the
config file as a part of `BeachballConfig`

Also improves the options types by making `CliOptions` extend shared
`RepoOptions`. Also sorts the options for easier readability in the future.

### DIFF
--- a/src/types/BeachballOptions.ts
+++ b/src/types/BeachballOptions.ts
@@ -5,87 +5,91 @@ import { ChangelogOptions } from './ChangelogOptions';
 
 export type BeachballOptions = CliOptions & RepoOptions & PackageOptions;
 
-export interface CliOptions {
+export interface CliOptions
+  extends Pick<
+    RepoOptions,
+    | 'access'
+    | 'branch'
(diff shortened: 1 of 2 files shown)


src/types/BeachballOptions.ts
```diff
@@ -5,87 +5,91 @@ import { ChangelogOptions } from './ChangelogOptions';
 
 export type BeachballOptions = CliOptions & RepoOptions & PackageOptions;
 
-export interface CliOptions {
+export interface CliOptions
```
ecraig12345, line 8: I like the alphabetizing but it makes diffing challenging--were any properties removed from here?

ling1726: No removals, apart from alphebetizing, the only change was to add `scope` to RepoOptions

PR #610, opened 2021-10-19
Scope down to .md and .json files only while merging publish branch

vreddi:


ecraig12345:
Closing because the PR is stale, and because it's not safe to assume by default that the user only wants these file types included. A better approach would be to add a config option to only include certain file types.

Issue #614, opened 2021-11-03
Option to specify git fetch/pull depth


abahuja:
currently beachball does a git pull before trying to push npm packages, but for larger repos, that fetch can take pretty long.
Is there a way we can specify the depth? or even better let beachball skip the fetch part?

ecraig12345:
Looks like this was fixed by #624

PR #624, opened 2022-01-05
Adding no fetch to publish

viditganpi:
This PR is expected to expose fetch flag and a depth option in CLI which will allow user to specify whether fetch is required in bump and when required what should be the depth of fetch. This is meant to address the issue https://github.com/microsoft/beachball/issues/614

### DIFF
--- a/src/publish/bumpAndPush.ts
+++ b/src/publish/bumpAndPush.ts
@@ -24,7 +24,11 @@ export async function bumpAndPush(bumpInfo: BumpInfo, publishBranch: string, opt
     // pull in latest from origin branch
     console.log('Fetching from remote');
     if(options.fetch !== false){
-      gitFailFast(['fetch', remote, remoteBranch], { cwd });
+      if(options.depth){
+        gitFailFast(['fetch', remote, remoteBranch, `--depth=${options.depth}`], { cwd });
+      }else{
+        gitFailFast(['fetch', remote, remoteBranch], { cwd });
+      }
     }
     const mergeResult = git(['merge', '-X', 'theirs', `${branch}`], { cwd });
     if (!mergeResult.success) {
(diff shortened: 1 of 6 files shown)


src/publish/bumpAndPush.ts
```diff
@@ -23,7 +23,13 @@ export async function bumpAndPush(bumpInfo: BumpInfo, publishBranch: string, opt
 
     // pull in latest from origin branch
     console.log('Fetching from remote');
-    gitFailFast(['fetch', remote, remoteBranch], { cwd });
+    if(options.fetch !== false){
+      if(options.depth){
+        gitFailFast(['fetch', remote, remoteBranch, `--depth=${options.depth}`], { cwd });
+      }else{
```
kenotron, line 29: this line has a weird spacing issue - can you make sure to do a prettier run?

kenotron:
```suggestion
      } else {
```

src/publish/bumpAndPush.ts
```diff
@@ -23,7 +23,13 @@ export async function bumpAndPush(bumpInfo: BumpInfo, publishBranch: string, opt
 
     // pull in latest from origin branch
     console.log('Fetching from remote');
-    gitFailFast(['fetch', remote, remoteBranch], { cwd });
+    if(options.fetch !== false){
```
kenotron, line 26:
```suggestion
    if (options.fetch !== false) {
```

src/publish/bumpAndPush.ts
```diff
@@ -23,7 +23,13 @@ export async function bumpAndPush(bumpInfo: BumpInfo, publishBranch: string, opt
...
+      if(options.depth){
```
kenotron, line 27:
```suggestion
      if (options.depth) {
```

kenotron (approved):
LGTM - thanks!

PR #633, opened 2022-01-29
Update the api call for prerelease to semver

PorterNan:
This PR is to fix the logic of prerelease tag not working.
According to api doc from semver, prerelease bump should be like this:
https://www.npmjs.com/package/semver#:~:text=semver.inc(%271.2.3%27%2C%20%27prerelease%27%2C%20%27beta%27)

This change will allow beachball to call semver using 'prerelease' parameter to do a correct prerelease bump

### DIFF
--- a/change/beachball-06d5d001-14ad-4767-bd88-1f59e62af52a.json
+++ b/change/beachball-06d5d001-14ad-4767-bd88-1f59e62af52a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update the api call for prerelease to semver",
+  "packageName": "beachball",
+  "email": "jiangnanhello@live.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


kenotron:
This fails a test - can you take a look?

awentzel:
> This fails a test - can you take a look?

@PorterNan anyway you can take a look at the failing test?

EisenbergEffect:
Super high priority for FAST and blocking time-sensitive critical releases. We need proper prerelease support please.

radium-v:
The failing test is `version bumping -> bumps all packages and uses prefix in the version` in `bump.test.ts` on line 607. The failing `expect` is on line 687.

awentzel:
Prerequisites include Node 12.  Only passes when executing `yarn run test` on Node `12.22.12`.

ecraig12345:
Looks like this was addressed by #658 instead.

PR #654, opened 2022-04-28
Add support for "workspace:" versions

### LINKED ISSUE #552: Feature: support "workspace:*" syntax to be replaced with the real versions at publish time.

dlannoye:
Resolves #552 and adds support for bumping and publishing of packages using dependencies with "workspace:" version prefixes. 

Since publishing only happens via npm which won't transform these during publish, workspace versions are temporarily transformed in package.json following the transformations laid out in the pnpm and yarn docs:
 * https://pnpm.io/workspaces#workspace-protocol-workspace
 * https://yarnpkg.com/features/workspaces#publishing-workspaces

For bumping workspace references like "workspace:*", "workspace:~", "workspace:^" are left as-is, while references like "workspace:~1.0.1" and "workspace:^1.0.1" will be updated to reflect the updated version of the dependency.


src/publish/performPublishOverrides.ts
```diff
@@ -0,0 +1,55 @@
+import { PackageInfos } from '../types/PackageInfo';
+import * as fs from 'fs-extra';
+
+export const acceptedKeys = ['types', 'typings', 'main', 'module', 'exports', 'repository', 'bin', 'browser', 'files'];
+
+export function performPublishOverrides(packagesToPublish: string[], packageInfos: PackageInfos) {
+  for (const pkgName of packagesToPublish) {
+    const info = packageInfos[pkgName];
+    const packageJson = fs.readJSONSync(info.packageJsonPath);
```
kenotron, line 9: Now that this function performs 2 separate actions, we should really document / separate it out so it's clearer what is done to the package.json files.

kenotron (dismissed):
I have a minor nit about documenting / refactoring of the performPublishOverrides()

kenotron (approved):

PR #656, opened 2022-05-04
Fix behavior of VersionGroupOptions.exclude

ecraig12345:
While working on another change, I noticed that the behavior of the `VersionGroupOptions.exclude` option was basically acting like a second `include` option: the passed patterns would have to be negated to work. This is very counterintuitive, not useful, and likely was a mistake.

Maybe this diff from one of the tests helps illustrate the before/after behavior. Before, `minimatch(packagePath, excludePattern)` must *return true* for one of the exclude patterns, so the exclude pattern had to be negated. After, the option behaves as expected.
```diff
  it('should return false if path is excluded with single exclude path', () => {
      // isPathIncluded(packagePath, include, exclude)
-    expect(isPathIncluded('packages/a', 'packages/*', '!packages/a')).toBeFalsy();
+    expect(isPathIncluded('packages/a', 'packages/*', 'packages/a')).toBeFalsy();
  });
```

This **could be a breaking change** if anyone was using `group.exclude`, but given the odd behavior, it seems less likely that anyone was relying on it.

(I also renamed the file the function is located in from `utils.ts` to `isPathIncluded` since it only has that one function.)


ecraig12345:
Too risky outside a major version

PR #659, opened 2022-05-16
Makes changes to the “CI Integration” page

AkimaLunar:
# Makes changes to the “CI Integration” page

1. Adds an Azure DevOps Example
2. Adds the `-y` flag to the CI script example

### DIFF
--- a/docs/concepts/ci-integration.md
+++ b/docs/concepts/ci-integration.md
@@ -57,7 +57,7 @@ Then inside the CI script, simply call `yarn publish:beachball` or `npm run publ
 
 ### With Azure DevOps
 
-**1 Npm authentication**. Azure DevOps has `npmAuthenticate@0` available to create a temporary npm auth token for publishing.
+**1 Npm authentication**. Azure DevOps has a pre-built task `npmAuthenticate@0` to create a temporary npm auth token for publishing.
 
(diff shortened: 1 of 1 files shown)


docs/concepts/ci-integration.md
```diff
@@ -36,16 +36,105 @@ A common requirement is to be able to publish to a private registry other than n
...
+- git.user
+- git.name
+- git.email
+
+Give your pipeline permissins to access these variables in _Pipeline permissions_.
+
+**2.1 Configure Git.** You can set up your Azure DevOps pipeline to add git credentials with Git CLI or you can extrapolate it in a Node script like this:
+
+```js
+const shell = require('shelljs');
```
kenotron, line 83: Why don't we give example in terms of the yaml pipeline itself? shelling out to a task in `just-scripts` seems overkill?

I'm thinking something like this:
```yaml
steps:
  - script: git config .... 
```

kenotron: You have most of it down below in the yaml example

AkimaLunar: 👌👌 Good point! I was kind of on a fence on this one. Ideally, I would want ADO to have a pre-built task for this.

PR #669, opened 2022-05-31
Fix link to change file

kwickramasekara:
This PR attempts to fix the "change file" broken link on [this page](https://microsoft.github.io/beachball/cli/change.html#change).

### DIFF
--- a/docs/cli/change.md
+++ b/docs/cli/change.md
@@ -6,7 +6,7 @@ category: doc
 
 # `change`
 
-This command walks you through a couple of questions and will generate the appropriate [change file](./concepts/change-files) in the `/change` directory. The generated file will be checked into the repo automatically for you. One of the niceties of using this utility to generate change files is that it will [check](./check) whether or not you even need a change file or not. Also, it will try to pull in recent commit messages to speed up change file generation.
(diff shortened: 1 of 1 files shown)


ecraig12345 (approved):
Thanks!

docs/cli/change.md
```diff
@@ -6,7 +6,7 @@ category: doc
 
 # `change`
 
-This command walks you through a couple of questions and will generate the appropriate [change file](./concepts/change-files) in the `/change` directory. The generated file will be checked into the repo automatically for you. One of the niceties of using this utility to generate change files is that it will [check](./check) whether or not you even need a change file or not. Also, it will try to pull in recent commit messages to speed up change file generation.
+This command walks you through a couple of questions and will generate the appropriate [change file](/concepts/change-files) in the `/change` directory. The generated file will be checked into the repo automatically for you. One of the niceties of using this utility to generate change files is that it will [check](./check) whether or not you even need a change file or not. Also, it will try to pull in recent commit messages to speed up change file generation.
```
ecraig12345, line 9:
```suggestion
This command walks you through a couple of questions and will generate the appropriate [change file](../concepts/change-files) in the `/change` directory. The generated file will be checked into the repo automatically for you. One of the niceties of using this utility to generate change files is that it will [check](./check) whether or not you even need a change file or not. Also, it will try to pull in recent commit messages to speed up change file generation.
```

ecraig12345 (changes requested):
Actually looking at this again, I think since the site is hosted at https://microsoft.github.io/beachball, the URL needs to start with `../` instead of `/`

ecraig12345 (approved):

PR #675, opened 2022-06-13
Update package-lock.json after bumping packages

### LINKED ISSUE #525: beachball not updating package-lock.json version field when publishing
Hello,

I am using beachball to manage a single-package repo (not a monorepo), which it says on [the beachball site](https://microsoft.github.io/beachball/): "single or monorepo". I'm using npm. I understand that the majority of monorepo users are using yarn, but npm is popular in the single repo ecosystem. Also, the best practice for npm is to use lockfiles (package-lock.json).

The issue is that package-lock.json files have a `version` field that corresponds to the `version` field of package.json, but beachball does not update this field in package-lock.json. Other tooling does, such as the `npm version` command. 

Currently our workaround is to do some manual script wrangling with git:
```sh
# Beachball doesn't update package-lock.json files when publishing.
- npx beachball publish --yes --no-push
- npm i --package-lock-only
- git add -A
- git commit -m "applying package updates [skip ci]"
- git push --set-upstream origin main
```
This is a bit uglier than the happy path:
```sh
npx beachball publish --yes --message "applying package updates [skip ci]"
```

This related issue (https://github.com/microsoft/beachball/issues/482) saw the same issue and added a postpublish hook, but it's not clear how to use it.

I do use beachball with yarn in other monorepos and it's wonderful. It would just be nice if beachball updated the lockfile.  Would this be hard?

ecraig12345:
If the package manager is detected to be npm (as determined by presence of `package-lock.json`), beachball should run `npm install --package-lock-only` to update `package-lock.json` after bumping or syncing.

I currently have this new behavior happening automatically rather than under a flag, because arguably it's something beachball should have been doing all along (the only reason it didn't was because the maintainers were primarily using beachball with yarn). If that seems risky, I could add a flag, but if we do that it should definitely become the default behavior in a future major version.

Fixes #525

### DIFF
--- a/src/bump/performBump.ts
+++ b/src/bump/performBump.ts
@@ -1,12 +1,15 @@
+import { spawnSync } from 'child_process';
+import fs from 'fs-extra';
+import os from 'os';
+import path from 'path';
 import { unlinkChangeFiles } from '../changefile/unlinkChangeFiles';
 import { writeChangelog } from '../changelog/writeChangelog';
-import fs from 'fs-extra';
-import path from 'path';
 import { BumpInfo } from '../types/BumpInfo';
 import { BeachballOptions, HooksOptions } from '../types/BeachballOptions';
 import { PackageDeps, PackageInfos } from '../types/PackageInfo';
+import { findProjectRoot } from '../paths';
 
-export function writePackageJson(modifiedPackages: Set<string>, packageInfos: PackageInfos) {
+export function writePackageJson(modifiedPackages: Set<string>, packageInfos: PackageInfos, cwd: string) {
   for (const pkgName of modifiedPackages) {
     const info = packageInfos[pkgName];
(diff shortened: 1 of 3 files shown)


src/bump/performBump.ts
```diff
@@ -44,7 +57,7 @@ export async function performBump(bumpInfo: BumpInfo, options: BeachballOptions)
 
   await callHook('prebump', bumpInfo, options);
 
-  writePackageJson(modifiedPackages, packageInfos);
+  writePackageJson(modifiedPackages, packageInfos, options.path);
```
kenotron, line 60: I would much prefer to have this be a separated function so we can dedicate "writePackageJson" to just writing the package.json. And then we can have a post-bump lock file update kind of function.

kenotron (commented):
a suggestion / nit to refactor so the writePackageJson is still the same.

kenotron (approved):

PR #677, opened 2022-06-23
Use path utilities from workspace-tools

ecraig12345:
Beachball has a file with various path utilities which are now redundant with utilities provided by workspace-tools (probably because they were copied from beachball into workspace-tools). This PR updates it to use the workspace-tools versions instead.

The only path utility I decided to keep in beachball is `getChangePath`, because that's definitely specific to beachball and not used elsewhere. (It exists in workspace-tools currently, but I have a PR https://github.com/microsoft/workspace-tools/pull/115 to remove it since it's really not appropriate for that package.)

Removal of the utilities is technically not a breaking change because they're not exported. If anybody was using the utilities via deep imports, they can switch to using the workspace-tools versions.

### DIFF
--- a/change/beachball-e8e41d79-466d-48e2-bc0d-91615011c641.json
+++ b/change/beachball-e8e41d79-466d-48e2-bc0d-91615011c641.json
@@ -1,6 +1,6 @@
 {
   "type": "minor",
-  "comment": "Use path utilities from workspace-tools",
+  "comment": "Use path utilities from workspace-tools and remove beachball's redundant implementations",
   "packageName": "beachball",
   "email": "elcraig@microsoft.com",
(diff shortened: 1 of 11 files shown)


change/beachball-e8e41d79-466d-48e2-bc0d-91615011c641.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Use path utilities from workspace-tools",
```
ecraig12345, line 3:
```suggestion
  "comment": "Use path utilities from workspace-tools and remove beachball's redundant implementations",
```

PR #685, opened 2022-07-13
Update workspace-tools and remove unused dependencies

ecraig12345:
Update `workspace-tools` to pick up an update of `git-url-parse` to address a security vulnerability. The new version unfortunately includes some [breaking changes](https://github.com/IonicaBizau/git-url-parse/releases/tag/12.0.0) which may impact consumers, so the `beachball` release is marked as a minor.

Also remove unused dependencies on `glob` and `git-url-parse`--I'm pretty sure this code moved to `workspace-tools` awhile back.


kenotron (approved):
LGTM!

PR #691, opened 2022-08-01
grouped packages propagate the change type based on ChangeType order

tannerlyons:
Previously, they were ordered based on the filesystem's sort order. The change type of the last change file in the array of changes  for a group would be propagated to all the change files in that group.

This PR figures out the max change type for the entire group and sets that into the `calculatedChangeTypes` for the packages in the group.

### DIFF
--- a/change/beachball-84084dc1-705a-499a-bf96-644bc4222a6f.json
+++ b/change/beachball-84084dc1-705a-499a-bf96-644bc4222a6f.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "grouped packages propagate the change type based on ChangeType order, not change file load order.",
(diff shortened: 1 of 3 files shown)


tannerlyons:
Hey team - it's been a while. Do we think any movement can happen on this MR or should we stop waiting? Are there people we can reach out to?

ecraig12345:
Hi, so sorry for the extreme delay in looking at this--I don't work on beachball regularly, but I'm still not sure how I missed this one for so long. I made another PR #1083 with a slightly different approach (calculating upfront in case of very large sets of changes).

If you're interested in contributing in the future (totally understandable if not), feel free to tag me, or email the address in the change files if there's no response.

PR #693, opened 2022-08-06
Reduce file reads and add verbose logs for determining changed packages

ecraig12345:
### Logging

I was having an issue when running beachball in a pipeline where it would generate change files for a bunch of packages that weren't actually changed, and I couldn't figure out why. So I made a custom build of beachball with additional verbose logs about determining changed files, and that helped me figure out the issue.

Since these logs could be helpful with others to similar issues, this PR adds them to official beachball under the `--verbose` option. (I know beachball could use an overall more coherent approach to logging, but I don't think that should block us from making incremental improvements like this.)

### Reducing file reads

While working on adding the verbose logs, I noticed that `getAllChangedPackages` was doing a lot of filesystem access (walking up the filesystem looking for package.json then reading it), with some attempt at caching, but the caching wouldn't work for out of scope or private packages. Also, the filesystem reads weren't even necessary since `packageInfos` contains all the info to match a file with a package and determine if the package should be published.

Changing approaches here revealed a bug with the `multiMonorepo` fixture: it was using the same name for packages in the different monorepos, which caused the packages from the second fixture monorepo to overwrite the first ones. Adding scopes to the different monorepos' package names fixes it.

### Options

There was an option `packageOptions.shouldPublish` that was documented on the website and being used in `getAllChangedPackages`, but it wasn't actually included in any of the types. I added it to the types and updated the docs to clarify its current behavior (I'd previously worked with someone who was confused by this): it can ONLY be used to *disable* publishing of a package that would otherwise be published, not to opt in a package that would otherwise be excluded.


src/changefile/getChangedPackages.ts
```diff
@@ -1,57 +1,96 @@
...
 import { BeachballOptions } from '../types/BeachballOptions';
-import { PackageInfos } from '../types/PackageInfo';
+import { PackageInfos, PackageInfo } from '../types/PackageInfo';
+
+function getMatchingPackageInfo(
+  file: string,
+  cwd: string,
+  packageInfosByPath: { [packageAbsPosixPath: string]: PackageInfo }
+) {
+  cwd = path.posix.normalize(cwd);
```
kenotron, line 16: This makes me wonder if we need to add windows to our github action os matrix

ecraig12345: Yeah that could be a good idea. The annoying thing is I don't even know if this is needed, I'm kind of just guessing it might be...

kenotron (approved):

Issue #714, opened 2022-08-08
Beacball publish process failure due to git fail fast


viditganpi:
The publish step of beachball has a few instances of `gitFailFast` which are not enclosed under any error handling, this leads to failure of the overall process in case `gitFailFast` command fails. 
For example: [Here](https://github.com/microsoft/beachball/blob/a74b1cc94d0f521755dc83a288b79a2a2845bc2a/src/commands/publish.ts#L98) branch deletion sometime results in error, thus failing the overall process. 
This can be handled for error and will fix the overall process.

ecraig12345:
Do you have any idea why the branch deletion is failing for you? It's only a local operation, so it should be very unlikely to fail.

viditganpi:
Not entirely sure, as there is no detailed log but have seen the overall process fail after the step and since there is no error handling for this my best guess is that something fails during deletion.

ecraig12345:
On failure, `gitFailFast` should log `CRITICAL ERROR: running git command` followed by the actual stdout/stderr, so any error should be visible in the logs.

PR #715, opened 2022-08-08
Adding error handling for publish

viditganpi:
Fixes https://github.com/microsoft/beachball/issues/714

### DIFF
--- a/change/beachball-ef09dabd-ce24-4044-af18-daf358a4f64c.json
+++ b/change/beachball-ef09dabd-ce24-4044-af18-daf358a4f64c.json
@@ -1,6 +1,6 @@
 {
   "type": "patch",
-  "comment": "Prettier correction",
+  "comment": "Moving non essential and retried commands to git from gitFailFast",
   "packageName": "beachball",
   "email": "viditganpi10@gmail.com",
   "dependentChangeType": "patch"
--- a/change/beachball-ef09dabd-ce24-4044-af18-daf358a4f64c.json
+++ b/change/beachball-ef09dabd-ce24-4044-af18-daf358a4f64c.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Prettier correction",
+  "packageName": "beachball",
+  "email": "viditganpi10@gmail.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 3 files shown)


change/beachball-ef09dabd-ce24-4044-af18-daf358a4f64c.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Prettier correction",
```
ecraig12345, line 3: This is used in the changelog, so it should describe the actual change.

src/commands/publish.ts
```diff
@@ -84,17 +84,21 @@ export async function publish(options: BeachballOptions) {
...
   const hasHash = currentHash !== null;
-  if (inBranch) {
-    console.log(`git checkout ${currentBranch}`);
-    gitFailFast(['checkout', currentBranch!], { cwd });
-  } else if (hasHash) {
-    console.log(`Looks like the repo was detached from a branch`);
-    console.log(`git checkout ${currentHash}`);
-    gitFailFast(['checkout', currentHash!], { cwd });
-  }
+  try {
```
ecraig12345, line 87: I don't think this is the right approach. `gitFailFast` is intended to be used with critical operations that are blocking if they fail. It also sets `process.exitCode = 1` even if you catch the error, so the step is likely to show up as failed in a CI build anyway.

A better approach would be to switch to using `git()` *only* for the branch deletion operation and logging a warning if the command failed. (The prior `git checkout` operations should continue to use `gitFailFast` because it's possible that the build could have later steps that will fail if it's not on the expected branch.)

ecraig12345 (changes requested):

src/commands/publish.ts
viditganpi: updated this, also updated similar case in bump and push

change/beachball-ef09dabd-ce24-4044-af18-daf358a4f64c.json
viditganpi: updated

ecraig12345 (approved):
Thanks for the updates!

PR #731, opened 2022-08-09
Update typescript to 4.3

ecraig12345:
Typescript 4.3 was released in May 2021 and is required by `ts-jest` 28 and probably an increasing number of other tools (also the latest `@types/jest` needs 4.1+). Especially since beachball is a command line tool where the primary use case doesn't involve types at all, upgrading to 4.3 seems fairly safe at this point.


kenotron:
Since this package isn't used as a library, we can actually use the latest typescript version if we wanted. Any reason why we stop at 4.3?

kenotron (approved):
just curious about 4.3 vs 4.5 vs 4.7

PR #734, opened 2022-08-10
Improve auth error handling and other logs during publishing

ecraig12345:
While debugging the issues with verdaccio that turned out to be due to npm 8 (somehow), I noticed that handling of logs during publishing could use some improvement. 

This PR improves the logging and some related things:
- Update the logs to be clearer in places
- Specifically check for auth errors during publishing, and doesn't retry if found (retries won't help in that case). Also don't try to publish any of the other packages since they'll presumably have the same issue.
- Reduce the number of separate `console.log`/`console.error` calls, mainly for convenient reading while testing
- Fix the types of the npm helper method
- Use the npm helper method everywhere (a couple places were still using `spawnSync` directly)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,6 +24,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
+      - run: npm i -g npm@6
       - run: yarn
       - run: yarn build
       - run: yarn docs:build
--- a/src/__e2e__/packageManager.test.ts
+++ b/src/__e2e__/packageManager.test.ts
@@ -23,8 +23,8 @@ describe('packageManager', () => {
       await registry.reset();
     });
 
-    it('can publish', () => {
-      const publishResult = packagePublish(testPackageInfo, registry.getUrl(), '', '');
+    it('can publish', async () => {
+      const publishResult = await packagePublish(testPackageInfo, registry.getUrl(), '', '');
       expect(publishResult.success).toBeTruthy();
 
       const showResult = npm(['--registry', registry.getUrl(), 'show', testPackageInfo.name, '--json']);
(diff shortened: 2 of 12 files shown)


src/__e2e__/packageManager.test.ts
```diff
@@ -23,8 +23,8 @@ describe('packageManager', () => {
       await registry.reset();
     });
 
-    it('can publish', () => {
-      const publishResult = packagePublish(testPackageInfo, registry.getUrl(), '', '');
+    it('can publish', async () => {
+      const publishResult = await packagePublish(testPackageInfo, registry.getUrl(), '', '');
```
ecraig12345, line 27: Most of the test changes are just because I made `packagePublish` return a promise, so now it needs to be awaited in the tests that use it

src/__e2e__/publishRegistry.test.ts
```diff
@@ -31,75 +33,9 @@ describe('publish command (registry)', () => {
       repositoryFactory.cleanUp();
       repositoryFactory = undefined;
     }
-    if (spy) {
-      spy.mockRestore();
-      spy = undefined;
-    }
-  });
-
-  it('will perform retries', async () => {
```
ecraig12345, line 40: This test is still here, but I moved it to the end--I think it makes more sense to test normal/successful publishing scenarios first

src/publish/publishToRegistry.ts
```diff
@@ -11,6 +11,8 @@ import { shouldPublishPackage } from './shouldPublishPackage';
 import { validatePackageDependencies } from './validatePackageDependencies';
 import { performPublishOverrides } from './performPublishOverrides';
 
+type Unpromisify<T> = T extends Promise<infer U> ? U : never;
```
ecraig12345, line 14: note, this could be changed to `Awaited` at some point once we update to TS where that's available (I forget which version)

.github/workflows/pr.yml
```diff
@@ -24,6 +24,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
+      - run: npm i -g npm@6
```
ecraig12345, line 27: Temporarily running tests with specifically npm 6, until the other PR fixing the tests goes in

kenotron (approved):
As always, amazing code *and* editing of my goofy English :)

PR #735, opened 2022-08-10
Fix issue with publishing tests in npm 8

ecraig12345:
Any e2e tests involving publishing to a mock verdaccio registry were failing for me locally, and I eventually figured out this was somehow due to a difference in behavior between npm 6 and 7/8: something about the newer versions messes up verdaccio's handling of anonymous access for publishing (even when it's enabled in the verdaccio config). 

Since I couldn't figure out the underlying issue and how to fix it properly (not for lack of trying...), I added a step in the mock registry setup to log in as a fake user. This works on both npm 6 and 8. (The `verdaccio-auth-memory` package is used to configure the fake user credentials.)

I also updated the PR build to test against npm 6 and 8, and updated the release build to specifically use npm 8.

Another issue I was sometimes having locally is that the mock registry setup couldn't find a free port. So I updated it to try 1000 ports instead of only 10.

Finally, while debugging all of this, I needed to see verdaccio logs. It doesn't really work to output the logs to the console due to the child process handling, so I added an environment variable VERDACCIO_LOG which will make it log to a file if set.

### DIFF
--- a/src/__fixtures__/registry.ts
+++ b/src/__fixtures__/registry.ts
@@ -1,6 +1,7 @@
-import { ChildProcess, spawn } from 'child_process';
+import execa from 'execa';
 // @ts-ignore
 import fp from 'find-free-port';
+import verdaccioUser from './verdaccioUser';
 
 const verdaccioApi = require.resolve('./verdaccio.js');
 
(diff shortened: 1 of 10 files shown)


kenotron (approved):

src/__fixtures__/registry.ts
```diff
@@ -1,6 +1,7 @@
-import { ChildProcess, spawn } from 'child_process';
+import execa from 'execa';
```
dzearing, line 1: what's this? I don't see it in package.json. maybe I'm blind.

ecraig12345: It's in the dependencies

dzearing: oops yep i see it.

PR #759, opened 2022-08-16
Error on duplicate package names

ecraig12345:
In a monorepo with multiple workspaces, it's theoretically possible that a package name could be reused across the different workspace roots (this happened with the multi-workspace fixture until I recently added scopes to the package names). This creates problems in `getPackageInfos` because the packages are stored in an object indexed by package name, and prior to this PR, it would silently overwrite the first package with the second. 

This PR makes beachball exit with an error instead. Arguably it could do just a console warning instead, but a package being entirely ignored (due to the duplicate name) is a bad enough situation that I think it's better to fail loudly and make the user fix it.

### DIFF
--- a/change/beachball-0b973948-5e88-4403-9731-6fbb9b375e4a.json
+++ b/change/beachball-0b973948-5e88-4403-9731-6fbb9b375e4a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Error on duplicate package names in different workspaces (for a monorepo with multiple workspaces)",
+  "packageName": "beachball",
+  "email": "elcraig@microsoft.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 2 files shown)


kenotron (approved):
Interesting - I wonder how yarn workspaces deal with this?!?!

ecraig12345:
> Interesting - I wonder how yarn workspaces deal with this?!?!

In this case the dupes are in different workspaces. This is roughly what it looked like in the test fixture previously (before I fixed the fixture):

```
repo-a/             
  package.json      workspaces defined here
  yarn.lock
  packages/
    pkg1/
      package.json  name: "pkg1"
repo-b/             
  package.json      workspaces defined here
  yarn.lock         
  packages/
    pkg1/
      package.json  name: "pkg1"
```

PR #762, opened 2022-08-17
Separate functional tests from E2E tests

ecraig12345:
Some of beachball's tests cover individual helper functions, but were being run as part of E2E tests because they need to use fixtures on the filesystem. The problem with this is that if one of the helper functions has a bug, that failure will probably also cause downstream failures in the actual E2E scenario tests, and then logs for the important failure in the helper function will be buried in logs for all the downstream failures.

This PR introduces a third category, functional tests, which are run after unit tests and before E2E tests. I moved some of the existing "E2E" tests into the functional group and added one new test.

NOTE: The files which show as deleted are not actually gone; they're just split into multiple new files.


src/__functional__/changefile/readChangeFiles.test.ts
```diff
@@ -0,0 +1,100 @@
+import _ from 'lodash';
```
ecraig12345, line 1: This was previously part of `__e2e__/changelog.test.ts`

src/__functional__/changefile/writeChangeFiles.test.ts
```diff
@@ -0,0 +1,111 @@
+import fs from 'fs-extra';
```
ecraig12345, line 1: This test is new

src/__functional__/changelog/writeChangelog.test.ts
```diff
@@ -0,0 +1,300 @@
+import { generateChangeFiles } from '../../__fixtures__/changeFiles';
```
ecraig12345, line 1: This was previously part of `__e2e__/changelog.test.ts`

src/__functional__/validation/areChangeFilesDeleted.test.ts
```diff
@@ -0,0 +1,54 @@
+import fs from 'fs-extra';
```
ecraig12345, line 1: This was previously part of `__e2e__/validation.test.ts`

src/__functional__/validation/isChangeFileNeeded.test.ts
```diff
@@ -0,0 +1,84 @@
+import { defaultRemoteBranchName, defaultRemoteName } from '../../__fixtures__/gitDefaults';
```
ecraig12345, line 1: This was previously part of `__e2e__/validation.test.ts`

kenotron (approved):

PR #765, opened 2022-08-22
Add github actions jest reporter and use jest projects

ecraig12345:
### Reporters

Enable the new github-actions jest reporter: https://jestjs.io/blog/2022/04/25/jest-28#github-actions-reporter

### Projects 

Switch to using Jest projects to run the different groups of tests (unit, functional, e2e). This means you can run `yarn test someTestName` to run just one test of any type. It also allows getting rid of the separate launch configurations for different test types.

The individual test group commands like `yarn test:e2e` have been updated to use `--selectProjects`, e.g. `yarn test --selectProjects e2e`. The downside of this is that `yarn test:unit someTestName` no longer works, but that's probably okay since now you can simply do `yarn test someTestName` instead.


kenotron (approved):

jest.config.js
```diff
@@ -1,9 +1,32 @@
-module.exports = {
+// @ts-check
+/** @type {import('@jest/types').Config.InitialOptions} */
+const comonOptions = {
```
ecraig12345, line 3: I tried specifying these options in the root object (so each project would only have `displayName` and `testMatch`), but that didn't seem to work. So I'm sharing them with object spread instead.

jest.config.js
```diff
@@ -1,9 +1,32 @@
-module.exports = {
+// @ts-check
+/** @type {import('@jest/types').Config.InitialProjectOptions} */
+const comonOptions = {
```
kenotron, line 3: Probably "commonOptions" ...
```suggestions
const commonOptions = {
```

jest.config.js
```diff
@@ -1,9 +1,32 @@
...
+/** @type {import('@jest/types').Config.InitialOptions} */
+module.exports = {
+  reporters: ['default', 'github-actions'],
   testTimeout: 60000,
-  testMatch: ['**/__tests__/**/*.test.ts'],
+  projects: [
+    {
+      displayName: 'unit',
+      testMatch: ['**/__tests__/**/*.test.ts'],
+      ...comonOptions,
```
kenotron, line 19: here: 
```suggestion
...commonOptions,
```

jest.config.js
```diff
@@ -1,9 +1,32 @@
...
+    },
+    {
+      displayName: 'functional',
+      testMatch: ['**/__functional__/**/*.test.ts'],
+      ...comonOptions,
```
kenotron, line 24: here: 
```suggestion
      ...commonOptions,
```

package.json
```diff
@@ -31,10 +31,11 @@
     "release": "node ./lib/cli.js publish -y",
     "release:docs": "yarn docs:build && yarn gh-pages -d docs/.vuepress/dist --dotfiles",
     "start": "tsc -w --preserveWatchOutput",
-    "test": "yarn test:unit && yarn test:func && yarn test:e2e",
-    "test:e2e": "jest --config jest.e2e.js",
-    "test:func": "jest --config jest.func.js",
-    "test:unit": "jest",
+    "test": "jest",
+    "test:all": "yarn test:unit && yarn test:func && yarn test:e2e",
```
kenotron, line 35: Can test:all not be just "jest" now too? Or do they have to STILL run in sequence?

src/__functional__/monorepo/getPackageInfos.test.ts
```diff
@@ -31,10 +33,11 @@ function cleanPackageInfos(root: string, packageInfos: PackageInfos) {
     // Also remove jest CLI or config options which get mixed in...
     for (const [key, value] of Object.entries(pkgInfo.combinedOptions)) {
       if (
-        (defaultOptions as any)[key] === value ||
+        _.isEqual(value, (defaultOptions as any)[key]) ||
+        _.isEqual(value, repoOptions[key]) ||
```
kenotron, line 37: Is this needed?

jest.config.js
kenotron: This is probably the best we can do!

package.json
ecraig12345: They don't strictly have to run in sequence, it's just easier to find the actual error that way. So you can run `yarn test` to run everything in whatever order jest decides, or `yarn test:all` to run in a friendlier order.

src/__functional__/monorepo/getPackageInfos.test.ts
```diff
@@ -1,13 +1,15 @@
...
 import { RepositoryFactory } from '../../__fixtures__/repositoryFactory';
 import { tmpdir } from '../../__fixtures__/tmpdir';
 import { getPackageInfos } from '../../monorepo/getPackageInfos';
 import { PackageInfos } from '../../types/PackageInfo';
 import { getDefaultOptions } from '../../options/getDefaultOptions';
 import { getCliOptions } from '../../options/getCliOptions';
-import { gitFailFast } from 'workspace-tools';
 
 const defaultOptions = getDefaultOptions();
+const repoOptions = require('../../../beachball.config');
```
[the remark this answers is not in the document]
ecraig12345: I got rid of this by making `getPackageOptions` ignore CLI and repo options if `process.env.NODE_ENV === 'test'`. That works at the moment because we don't have test coverage for any scenarios where the packages would need to inherit options from the fixture...

kenotron (approved):

PR #778, opened 2022-09-20
feat: Adds precommit hook

ling1726:
Adds a precommit hook that is called before bumps are committed and pushed. Allows users to modify or create files based on the entire release instead of per package.

### DIFF
--- a/src/publish/mergePublishBranch.ts
+++ b/src/publish/mergePublishBranch.ts
@@ -31,9 +31,7 @@ export async function mergePublishBranch(
   return result!;
 }
 
-/**
- * Calls a specified hook for each package being bumped
- */
+/** Calls the `precommit` hook specified in `options` */
 async function precommitHook(options: BeachballOptions) {
   const hook = options.hooks?.precommit;
   if (!hook) {
--- a/src/publish/mergePublishBranch.ts
+++ b/src/publish/mergePublishBranch.ts
@@ -1,13 +1,22 @@
-import { git } from 'workspace-tools';
+import { git, GitProcessOutput } from 'workspace-tools';
+import { BeachballOptions } from '../types/BeachballOptions';
 
-export function mergePublishBranch(publishBranch: string, branch: string, message: string, cwd: string) {
-  let result: ReturnType<typeof git>;
+export async function mergePublishBranch(
+  publishBranch: string,
+  branch: string,
+  message: string,
+  cwd: string,
+  options: BeachballOptions
+) {
+  let result: GitProcessOutput;
   let mergeSteps = [
     ['add', '.'],
     ['commit', '-m', message],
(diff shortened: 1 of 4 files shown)


Hotell:
this is needed for https://github.com/microsoft/fluentui/issues/24207

src/publish/mergePublishBranch.ts
```diff
@@ -21,3 +30,18 @@ export function mergePublishBranch(publishBranch: string, branch: string, messag
...
+async function precommitHook(options: BeachballOptions) {
+  const hook = options.hooks?.precommit;
+  if (!hook) {
+    return;
+  }
+
+  const hookRet = hook(options.path);
+  if (hookRet instanceof Promise) {
+    await hookRet;
+  }
```
ecraig12345, line 46: You can await this even if it doesn't return a promise.
```suggestion
  await hook(options.path);
```

src/publish/mergePublishBranch.ts
```diff
@@ -1,14 +1,23 @@
...
+) {
+  let result: GitProcessOutput;
   let mergeSteps = [
     ['add', '.'],
     ['commit', '-m', message],
     ['checkout', branch],
     ['merge', '-X', 'ours', publishBranch],
   ];
 
+  await precommitHook(options);
```
ecraig12345, line 19: Would you mind moving this before the `mergeSteps` and `result`? On initially skimming this I was a bit confused by seeing the commands such as git add before the hook (even though on more careful reading they're not run until after).

ecraig12345 (dismissed):

src/types/BeachballOptions.ts
```diff
@@ -153,6 +153,11 @@ export interface HooksOptions {
    * to the filesystem. May be called multiple times during publish.
    */
   postbump?: (packagePath: string, name: string, version: string) => void | Promise<void>;
+
+  /**
+   * Runs once after all bumps to all packages before commiting changes
```
ecraig12345, line 158:
```suggestion
   * Runs once after all bumps to all packages before committing changes
```

ecraig12345 (dismissed):

src/publish/mergePublishBranch.ts
```diff
@@ -21,3 +30,15 @@ export function mergePublishBranch(publishBranch: string, branch: string, messag
   }
   return result!;
 }
+
+/**
+ * Calls a specified hook for each package being bumped
+ */
```
ecraig12345, line 36: Just realized this comment is probably an incorrect copy/paste
```suggestion
```

ling1726: uuuh no that's correct ?

ecraig12345 (approved):

PR #780, opened 2022-09-21
Fix handling of shallow clones when checking for change files

ecraig12345:
By default, [github `actions/checkout`](https://github.com/actions/checkout) only fetches the current merge ref of the PR (no history and no other branches). This causes problems for `beachball check`, because it doesn't have all the history that it needs to compare against.

These problems have existed for quite some time, in any repo using `actions/checkout` > v1 without `fetch-depth: 0`. And the `workspace-tools` git changed files helpers were hiding the issue by returning empty arrays if git commands failed.

Fix is to add explicit checks on the beachball side for various conditions related to missing history. If history is missing and fetching is enabled, it will fetch the extra history. If fetching is disabled, it will print informative error messages.

(You can see that it works because this PR's initial build includes the log about unshallowing, and ultimately failed due to a missing change file!)

### DIFF
--- a/src/changefile/getChangedPackages.ts
+++ b/src/changefile/getChangedPackages.ts
@@ -8,6 +8,84 @@ import { getScopedPackages } from '../monorepo/getScopedPackages';
 import { BeachballOptions } from '../types/BeachballOptions';
 import { PackageInfos, PackageInfo } from '../types/PackageInfo';
 
+/**
+ * Ensure that adequate history is available to check for changes between HEAD and `options.branch`.
+ * Otherwise attempting to get changes will fail with an error "no merge base".
+ */
+function ensureHistory(options: BeachballOptions) {
+  const { fetch, path: cwd, branch } = options;
+  const { remote, remoteBranch } = parseRemoteBranch(branch);
+
+  const fetchMitigationSteps = `- Omit the "--no-fetch" / "--fetch=false" option from the command line
+- Remove "fetch: false" from the beachball config
+- If this is a CI build, ensure that adequate history is being fetched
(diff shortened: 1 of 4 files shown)


src/changefile/getChangedPackages.ts
```diff
@@ -3,11 +3,83 @@ import path from 'path';
...
+  const hasCommonCommit = git(['merge-base', branch, 'HEAD'], { cwd }).success;
+  if (!hasCommonCommit) {
+    // This might be a shallow repo, and it's necessary to unshallow the head branch for comparison
+    const isShallow = git(['rev-parse', '--is-shallow-repository'], { cwd }).stdout.trim() === 'true';
+    if (isShallow) {
+      if (fetch) {
+        // Fetch more history (if needed, this could be optimized later to only deepen by e.g. 100 commits at a time)
+        try {
+          console.log('This is a shallow clone. Unshallowing to check for changes...');
+          git(['fetch', '--unshallow'], { cwd, throwOnError: true });
```
kenotron, line 56: Can you confirm this "throwOnError" flag? I can't find doc for it: https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options

ecraig12345: Ah it's added in https://github.com/microsoft/workspace-tools/pull/193 which isn't actually merged yet (I was testing this branch against a tgz of those changes at first)

ecraig12345: I got rid of the part that required the workspace-tools update

kenotron (approved):

PR #783, opened 2022-09-22
Fix commit message suggestions in change command

### LINKED ISSUE #781: List of suggested change messages is empty
When running `beachball change`, recent commit messages are no longer suggested.

ecraig12345:
(Currently also includes changes from #782)

`prompts` doesn't support "freeform"/custom input on autocomplete prompts (https://github.com/terkelg/prompts/issues/131). Since freeform input is sometimes desirable for change descriptions, we had a hack implemented to make it work: making a custom list of suggestions with the current input added at the end.

Something in [`prompts@2.2.0`](https://github.com/terkelg/prompts/releases/tag/v2.2.0) broke that hack. This PR implements a new hack (suggested in the issue) to modify the internal state of the prompt instead...which is not ideal for obvious reasons, but the alternative appears to be removing suggestions entirely or not allowing freeform input, neither of which are good.

(While I was working on this, I also implemented case-insensitive filtering of suggestions.)

Fixes #781


kenotron (approved):

Bryan-Cee:
fixes #784

PR #791, opened 2022-10-28
Support for custom tag separator

liby:
## Motivation

I noticed that the hyphen in the current `Tag` is hard code. E.g: `foo_v1.0.1`. This is likely to conflict with the `tag` format of some repositories.  E.g: `foo@v1.0.1`.


ecraig12345 (changes requested):
The character `_` is called an underscore, not a hyphen. A better (more generic) name for the option, if it's needed at all, would probably be `gitTagSeparator`.

I'm also not quite understanding the need for this option. Can you please provide more details about your scenario?

liby:
> I'm also not quite understanding the need for this option. Can you please provide more details about your scenario ?

Many existing projects may use a tag format different from `projectName_v1.0.1`, for example, the format used for [changesets](https://github.com/changesets/changesets) is `@changesets/write@0.2.2, @changesets/cli@2.25.2`...

If there is no flexible configuration, then the introduction of beachball in many existing projects is worth considering, as it is likely to result in inconsistent formatting of `tag`

PR #792, opened 2022-10-28
Support customize change files commit 

### LINKED ISSUE #307: Personalize change files commit message
https://github.com/microsoft/beachball/blob/ca6a8b2ed1675a533a7d62c8b55de687bc7ba40a/packages/beachball/src/changefile/writeChangeFiles.ts#L47

Would be great if we could personalize it. For example, today I'm using commitzen as our conversion and it breaks our pre-hook commit message lint.

liby:
Fix #307


src/changefile/writeChangeFiles.ts
```diff
@@ -13,12 +13,14 @@ export function writeChangeFiles({
   changes,
   cwd,
   commitChangeFiles = true,
+  commitMessage = 'Change files',
```
ecraig12345, line 16: This default should be supplied in `getDefaultOptions` rather than here.

src/types/BeachballOptions.ts
```diff
@@ -57,6 +57,8 @@ export interface RepoOptions {
   /** Bump dependent packages during publish (bump A if A depends on B) */
   bumpDeps: boolean;
   changeFilePrompt?: ChangeFilePromptOptions;
+  /** Customize change files commit message */
+  changeFilesCommitMessage?: string;
```
ecraig12345, line 61: I think it's okay for this to just be called `commitMessage`

ecraig12345 (dismissed):

ecraig12345:
Looks like the build is failing because the new parameter is missing from a couple tests.

kwickramasekara:
Bump - wanting to use this change and would greatly appreciate it if this could be reviewed.

elpatronaco:
Why was this closed?

PR #794, opened 2022-10-29
feat: skips the prompts for publish on CI

liby:
This change is intended to eliminate the need to pass the `--yes` option when running beachball in a CI/CD integrated environment (not specifically noted in the documentation).

cc @ecraig12345


ecraig12345:
I fixed this in a different way in #835.

PR #797, opened 2022-11-07
ensure that all packages bumped get a new entry in the changelog.json…

LarryKiniu:
This PR fixes an issue in beachball where, If package **_foo_** depends on package **_bar_** with an entry in it's _package.json_:
``
"bar":"*"
``
If you update **_bar_**. both packages should be bumped but only **_bar_** is getting a new entry in it's _changelog.json_ file.

Each package, should get a new entry in it's _changelog.json_ file.

### DIFF
--- a/src/bump/setDependentVersions.ts
+++ b/src/bump/setDependentVersions.ts
@@ -24,7 +24,7 @@ export function setDependentVersions(
           if (packageInfo) {
             const existingVersionRange = deps[dep];
             const bumpedVersionRange = bumpMinSemverRange(packageInfo.version, existingVersionRange);
-            if (existingVersionRange !== bumpedVersionRange) {
+            if (bumpedVersionRange === "*" || existingVersionRange !== bumpedVersionRange) {
               deps[dep] = bumpedVersionRange;
 
               dependentChangedBy[pkgName] = dependentChangedBy[pkgName] || new Set<string>();


evancharlton:
This feels like something that should be covered by unit tests, as this could be easily overlooked.

LarryKiniu:
> This feels like something that should be covered by unit tests, as this could be easily overlooked.

I can work on adding some unit tests for this scenario.

kenotron (approved):

ecraig12345:
@LarryKiniu You need to run `yarn build` then `yarn change` to generate change files.

Netail:
Seems to also add "bumped package to vX.X.X" of packages which didn't got bumped.

E.g if you have 4 packages; 

```json
{
    "name": "foo",
    "version": "0.1.0",
    "dependencies": {
        "bar": "*",
        "baz": "*"
    }
}
```
```json
{
    "name": "bar",
    "version": "0.1.0",
}
```
```json
{
    "name": "baz",
    "version": "0.1.0",
}
```
```json
{
    "name": "faa",
    "version": "0.1.0",
    "baz": "*"
}
```
If you have a changefile for Bar, it will update the changelog for all.
Foo's changelog will include; "bump bar to X" & "bump baz to X", but also Faa's changelog will be bumped with the baz package, while this one wasn't included in the changefile

PR #815, opened 2023-01-06
Ignore scripts when bumping packages.

### LINKED ISSUE #814: Beachball should ignore scripts when running bump
When running Beachball publish we noticed that our postinstall scripts were running (in our case build and lint) which greatly increased the time publish took. We should ignore scripts when beachball updates package-lock.json instead.

brandonthomas:
Since Beachball bump calls NPM I to update package-lock.json, it also currently runs scripts. This isn't needed or desired so I've updated bump to ignore scripts.

Fixes #814


mltejera (approved):

ecraig12345 (approved):

change/beachball-fc620083-adfb-49f8-9d54-833bef094424.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Ignore postinstall scripts when running npm install after bumping packages",
```
ecraig12345, line 3:
```suggestion
  "comment": "Use `--ignore-scripts` when running `npm install` after bumping packages (for repos using npm only)",
```

ecraig12345 (approved):

Issue #826, opened 2023-02-09
Push failure when file not found


viditganpi:
With the current flow beachball fetches from remote before doing the bump and push for the versions, here it can happen that one of the packages meant to be bumped is deleted from main, this causes the whole process to fail due to `package.json` for that package not being found.

ecraig12345:
Fixed by #827

PR #827, opened 2023-02-09
Handle file deletion scenario

viditganpi:
Fixes issue https://github.com/microsoft/beachball/issues/826

### DIFF
--- a/change/beachball-8b09ffe5-dd69-4c08-b86d-1c764a046c97.json
+++ b/change/beachball-8b09ffe5-dd69-4c08-b86d-1c764a046c97.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Handle file not found issue",
+  "packageName": "beachball",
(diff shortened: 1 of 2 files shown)


ecraig12345 (changes requested):
The issue here would happen if this package.json doesn't exist after the merge, right? If so, I think it's better to explicitly check `if (!fs.existsSync(info.packageJsonPath))` before attempting the update, and if not found, log a warning that the package appears to have been deleted between the build start and the bump attempt.

I was also going through what beachball does during publish to try and understand whether this is even a semi-valid scenario, or if the whole publish attempt should fail (or if we should do something else to detect this case earlier)--honestly I'm still a bit confused after reading the code 🙄 but I'm pretty sure that at this point in the process, when things have already been published to npm, it's "less bad" to log a warning rather than failing the build.

viditganpi:
made the requested changes

ecraig12345 (approved):

PR #841, opened 2023-03-20
Add optional extra logging for npm publish

ecraig12345:
Currently there's no way to make beachball show the actual list of files published with `npm publish` due to the `--loglevel warn` arg (the list of files uses `notice` level).

This PR makes Beachball's `--verbose` option enable `--loglevel notice` for `npm publish` and pipe output to the console, rather than only showing output if there's an error.

(If someone needs even more detailed `npm publish` logs in the future, we could add a separate option for more detailed control of npm loglevel, but I think enabling basic output under `--verbose` should be good enough for most situations.)


ecraig12345:
Never mind, this is less useful than I thought...

Issue #844, opened 2023-04-11
Beachball updating package.json version for private package


viditganpi:
Currently beachball collects package info at the start of publish process and it might use a pre-fetched remote repo version, and post publish it performs the package bump, till this point it can happen that a manual update to package.json has been pushed for a private package. 

Now this update is actually ignored, and the info collected earlier is used to update the package.json version during the bump process and even for private package this step is not skipped. 

Ideally for private package package.json version update should not happen.

PR #845, opened 2023-04-11
Handle private package version update

viditganpi:
Fixes https://github.com/microsoft/beachball/issues/844

### DIFF
--- a/change/beachball-8dfc65fb-0740-4b07-b1d3-ce89e06bfbf0.json
+++ b/change/beachball-8dfc65fb-0740-4b07-b1d3-ce89e06bfbf0.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Handling package version update for private package",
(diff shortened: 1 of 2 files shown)


change/beachball-8dfc65fb-0740-4b07-b1d3-ce89e06bfbf0.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Handling package version update for private package",
```
ecraig12345, line 3:
```suggestion
  "comment": "When bumping, don't modify the version of private packages",
```

ecraig12345 (approved):
Approving because from looking at the other bumping code, I *think* the intention was that private packages should *never* be bumped on general principle (the issue you highlighted about conflicts is more of an edge case), and the fact that the version was being updated is just an implementation artifact.

PR #863, opened 2023-05-04
Speed up CI

ecraig12345:
~~`yarn install` is super slow on Windows CI machines, so try using midgard-yarn to speed it up.~~

That didn't help...trying using [`actions/setup-node`](https://github.com/actions/setup-node)'s `cache` option instead?


ecraig12345:
This might be a small improvement, but it seems like the slowness has mainly moved to unpacking the .tgz file of the cache...

PR #876, opened 2023-05-17
Improve logging during publishing

ecraig12345:
Add more logs (and more line breaks) and reformat/group some existing logs for improved readability. Specifically for various package validation steps, it's easier to read the logs if the successful and error or skipped packages are grouped together as a list under a heading rather than interleaved or printed with lots of repeated text.

### DIFF
--- a/src/publish/publishToRegistry.ts
+++ b/src/publish/publishToRegistry.ts
@@ -10,6 +10,7 @@ import { toposortPackages } from './toposortPackages';
 import { shouldPublishPackage } from './shouldPublishPackage';
 import { validatePackageDependencies } from './validatePackageDependencies';
 import { performPublishOverrides } from './performPublishOverrides';
+import { formatList } from '../logging/format';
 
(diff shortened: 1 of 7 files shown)


src/publish/publishToRegistry.ts
```diff
@@ -53,29 +61,19 @@ export async function publishToRegistry(originalBumpInfo: BumpInfo, options: Bea
...
-        await maybeAwait;
-      }
+      await prepublishHook(path.dirname(packageInfo.packageJsonPath), packageInfo.name, packageInfo.version);
     }
   }
 
   // finally pass through doing the actual npm publish command
-  let hasAuthError = false;
-  for (const pkg of packagesToPublish) {
-    if (hasAuthError) {
```
ecraig12345, line 70: This separate hasAuthError check appears to be unnecessary since the end of this `for` loop throws if individual package publishing was unsuccessful for any reason

PR #896, opened 2023-07-17
Move workspace publish overrides after prepublish hook execution

tronguye:
We are trying to have our yarn 1 repo be able to:
1. Locally, refer to local workspace package version
2. At publish time, change to ^x.y.z, where x.y.z is the post-bump version

To accomplish this, we are hoping to use the prepublish hook to change the internal packages to workspace:^ version, which Beachball already has logic to replace with ^x.y.z. Locally, we'd keep a specific version such as * to resolve to local workspace version.

We attempted this with the postbump hook, but postbump hook is also executed in the git commit/push route and thus pushed workspace:^ versions to our upstream (which yarn 1 can't resolve). As such, we propose that we can execute the prepublish hook before the workspace range overrides. This will allow us to hook into the publish, change to workspace:^ for internal dependencies, which is then handed back off to Beachball to override to the right version to publish

### DIFF
--- a/change/beachball-dcfa4985-8bb6-49b8-85db-365302dd9325.json
+++ b/change/beachball-dcfa4985-8bb6-49b8-85db-365302dd9325.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Move workspace publish overrides after prepublish hook execution",
+  "packageName": "beachball",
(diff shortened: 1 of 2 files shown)


tronguye:
Closing pull request, talked to Elizabeth in private and we agreed that this doesn't match the semantics of "prepublish"

PR #900, opened 2023-07-20
Pass packageInfos into prepublish, postpublish, and postbump hooks

tronguye:
Same end goal as #896 

We are trying to have our yarn 1 repo be able to:

Locally, refer to local workspace package version using the * version
At publish time, change to ^x.y.z, where x.y.z is the post-bump version

To accomplish this, we are hoping to use the prepublish hook to change the internal packages to ^x.y.z. Passing the post-bump versions to the prepublish hook would allow for this to happen without file i/o on our end

### DIFF
--- a/src/bump/callHook.ts
+++ b/src/bump/callHook.ts
@@ -14,15 +14,10 @@ export async function callHook(
     return;
   }
 
-  const bumpVersions = Object.keys(packageInfos).reduce((accumulator, pkgName) => {
-    accumulator[pkgName] = packageInfos[pkgName].version;
-    return accumulator;
-  }, {} as Record<string, string>);
-
   for (const pkg of affectedPackages) {
     const packageInfo = packageInfos[pkg];
     const packagePath = path.dirname(packageInfo.packageJsonPath);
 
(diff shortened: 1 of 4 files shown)


src/bump/callHook.ts
```diff
@@ -16,6 +16,8 @@ export async function callHook(
 
   for (const pkg of affectedPackages) {
     const packageInfo = packageInfos[pkg];
-    await hook(path.dirname(packageInfo.packageJsonPath), packageInfo.name, packageInfo.version);
+    const packagePath = path.dirname(packageInfo.packageJsonPath);
+
+    await hook(packagePath, packageInfo.name, packageInfo.version, packageInfos);
```
tronguye, line 21: @ecraig12345 Should we clone package infos before passing? We're susceptible to side effects by hooks in the current calling pattern

tronguye: Marked param as Readonly<> in the typing of hooks. Doesn't technically stop a consumer from casting as writable and causing side effects. lmk what you think

ecraig12345 (approved):

PR #938, opened 2024-01-16
Adding support for identifier base

JulienDesalliers:
[Pre-release identifier](https://docs.npmjs.com/cli/v6/using-npm/semver#prerelease-identifiers) in semver allows to control the identifier base when bumping a pre-release package.

With beachball, the identifier base is always set to zero; the goal of this PR is to support zero-based, one-based and no identifier base.

The options for the identifier base are "0", "1" and false.

Here is how it works with semver: 
image

### DIFF
--- a/src/types/BeachballOptions.ts
+++ b/src/types/BeachballOptions.ts
@@ -3,6 +3,7 @@ import { ChangeInfo, ChangeInfoMultiple, ChangeType } from './ChangeInfo';
 import { ChangeFilePromptOptions } from './ChangeFilePrompt';
 import { ChangelogOptions } from './ChangelogOptions';
 import { PackageInfos } from './PackageInfo';
+import type { inc } from 'semver';
 
 export type BeachballOptions = CliOptions & RepoOptions & PackageOptions;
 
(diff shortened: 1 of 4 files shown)


src/types/BeachballOptions.ts
```diff
@@ -127,6 +128,11 @@ export interface RepoOptions {
   path: string;
   /** Prerelease prefix for packages that are specified to receive a prerelease bump */
   prereleasePrefix?: string | null;
+  /** This is for prerelease. Set it to "0" for zero-based or "1" for one-based.
+   *  Set it to false to omit the prerelease number.
+   *  @default "0"
+   */
+  identifierBase?: inc.IdentifierBase | false;
```
ecraig12345, line 135: To avoid a type reference to semver, can you change this to `'0' | '1' | false`? The type reference might not resolve in consumer repos if they're using types in their beachball config.

JulienDesalliers: Makes sense. Done!

ecraig12345 (approved):

PR #943, opened 2024-02-21
Add option to include private packages for changefile and bump

kjellknapen:
For some of our monorepo projects we currently version our packages with lerna version. We are making a switch to microsoft beachball for more control. Some of our packages are non JS based but do have a package.json. These packages need to bumped to the correct version but don't have to be published.

Their package.json contains `"private": true`. As of right now beachball didn't contain an option to allow the bumping of private packages so with this PR I added the config option to allow just that.

the new option ìncludePrivatePackages` defaults to false and is only used during the changeset/bump commands. This is hopefully something other people can benefit from.

### DIFF
--- a/docs/overview/configuration.md
+++ b/docs/overview/configuration.md
@@ -55,32 +55,33 @@ It's also common to have a repo-level `beachball.config.js` and any individual p
 
 For the latest full list of supported options, see `RepoOptions` [in this file](https://github.com/microsoft/beachball/blob/master/src/types/BeachballOptions.ts).
 
-| Option                  | Type                                     | Default           | Option Type          | Description                                                                                     |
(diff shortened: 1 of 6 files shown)


ecraig12345 (changes requested):
Thanks for the contribution, but there are several other places in the beachball logic which would require updates to fully handle private packages, and I'm not sure if it's a feature we want. What's the scenario in your repo that requires private package versions to be bumped?

timmevandermeer:
@ecraig12345 What about support for disabling `publish` per package? Currently it seems that if you specify in a nested beachball config file that for that package `publish: false` it gets ignored as all packages in `bumpInfo` get published. I have some packages I do want to version and keep a changelog for, but don't want to publish.

PR #944, opened 2024-02-21
Add release types for major and minor prereleases

kjellknapen:
Allow user to do prereleases of major and minor releases. It's very weird that isn't a possibility right now.


ecraig12345:
This makes sense but would probably be perceived as a breaking change in a lot of repos that use beachball today, since suddenly several new change types would be available and have to be manually added to their disallowedChangeTypes if not desired. I made an issue #947 to track adding this whenever we do a new major version (no specific plans for timing yet). 

In the meantime, I think a workaround would be to revert the `promptForChange` additions, but keep the new `SortedChangeTypes`. This would allow the new types to be specified as `--type` on the command line (or by editing the change file) and ensure they're handled with proper precedence. And then if you wanted to include the types in your repo's prompt, you could implement a custom `changeFilePrompt` in beachball.config.js.

kjellknapen:
@ecraig12345 Alright, I removed the other pre* versions from the prompt. Is that something that can be added in a minor/patch release in this way or do I need to make other changes to ensure we don't break current implementation?

kjellknapen:
I could optionally also disallow them all if prerelease is disallowed, but that may be a hacky solution?

ecraig12345:
@kjellknapen I think it's fine if new change types are allowed by default as long as they don't show up in the prompt.

I ran a build and it looks like you'll also need to update `src/changelog/renderPackageChangelog.ts` `groupNames` with the new types. Those are used as headers in the CHANGELOG.md if any changes of that type are present.

kjellknapen:
@ecraig12345 Done! Thanks for the quick response. I used the same title as the respective versions but added (pre-release)  behind them.

ecraig12345:
Looks like you accidentally installed with pnpm (this repo uses yarn)--can you delete the pnpm lock file?

kjellknapen:
Oops my bad, was working in a different project with pnpm, so I was in that flow. I removed it

ecraig12345 (dismissed):

ecraig12345:
Oh I guess you need a change file too.

ecraig12345 (dismissed):

kjellknapen:
I guess I need to update some tests aswell

src/__tests__/changefile/changeTypes.test.ts
kjellknapen, line 1: @ecraig12345  Does the test failing here signal that my changes are still breaking?

ecraig12345 (approved):

PR #946, opened 2024-02-26
Add context on current running command for change file transforms

christiango:
Beachball has a nice feature that lets you run custom transformations on change files. Our team is using them today to augment beachball changelogs with links to the ADO PRs corresponding to the release.

That being said, in the current implementation we are having a negative impact on local beachball change times. This is because beachball runs this transform during commands like change and publish. For our particular transform, we don't want to run it during yarn change, since that runs before we have a merged PR.

With this change, the custom beachball change file transform can decide to skip the expensive step to compute the PR URL if we are not running in the context of a beachball publish

### DIFF
--- a/change/beachball-a311609d-38f8-420d-8cd7-a5a98e5876ba.json
+++ b/change/beachball-a311609d-38f8-420d-8cd7-a5a98e5876ba.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Add context on the current running command for changeFiles transform",
+  "packageName": "beachball",
+  "email": "1581488+christiango@users.noreply.github.com",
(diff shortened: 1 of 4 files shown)


change/beachball-a311609d-38f8-420d-8cd7-a5a98e5876ba.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
ecraig12345, line 2: Should be minor since it's a new API, in some sense
```suggestion
  "type": "minor",
```

ecraig12345 (approved):

ecraig12345 (approved):

PR #957, opened 2024-04-09
Ensure it bumps premajor, preminor and prepatch with a prereleasePrefix

kjellknapen:
My last PR allowed us to add the options for new PrereleaseTypes: Premajor, preminor and prepatch. But the if statement in the current code overruled these types when the prereleasePrefix is checked.

In my opinion this check should change since having this means you can't add the prerelease prefix to the config cause it would handle every change as a prerelease. But since changing that would be a breaking change I added a different if check that should keep working in the current use cases


kjellknapen:
@ecraig12345 Noticed an issue in my previous PR

ecraig12345 (approved):

Issue #970, opened 2024-08-08
Feature Request: Option to prevent auto bumping peerDependencies


SevenOutman:
## The Problem

Hi team, thanks for making this tool! It's been helpful to keep our packages up-to-date in the monorepo.

One thing that turns out inconvenient is that, beachball does not seem to handle `peerDependencies` properly, by automatically bumping `peerDependencies` version when new versions are published.

For example, there is a `base` package and a `feature` package in the monorepo, and `base` is a `peerDependencies` of `feature`.

```json
{
  "name": "@org/feature",
  "version": "1.0.0,
  "peerDependencies": {
    "@org/base": "^1.0.0"
  }
}
```

When publishing a new version of `@org/base`, say `1.1.0`, beachball would update the `feature`'s package.json to depended on the new version of `base`.

```diff
{
  "name": "@org/feature",
- "version": "1.0.0,
+ "version": "1.0.1,
  "peerDependencies": {
-   "@org/base": "^1.0.0"
+   "@org/base": "^1.1.0"
  }
}
```

This is not expected. Unlike `dependencies`, bumping `peerDependencies` version means that `feature` no longer works with the older version of `base`, which is both false and breaking.

## This Proposal

Can we have an option `--bump-peer-deps` for `beachball bump` and `beachball publish` that, when set to `false`, prevents beachball from bumping the `peerDependencies` versions.

Any thoughts?

jvince:
@SevenOutman This would make sense because your @org/base as peer dependency accepts minor version changes. Have you tried changing the dependency range from `^1.0.0` to `~1.0.0`?

SevenOutman:
Thanks @jvince for your response! However I don't think this works for me.

> Have you tried changing the dependency range from ^1.0.0 to ~1.0.0?

With this, when I publish base 1.0.1, I will still have the peer dependencies bumped, right?

My point was that **peer dependencies should not be auto bumped** at all. Because its semantic is _significantly_ different from that of dependencies/devDependencies - for which auto-bumping is fine. 
Increasing the version of a peer dependency declaration itself is a breaking change (for reference https://github.com/semver/semver/issues/502#issuecomment-466501843), because it means my packages stops working with the previous version of that dependency. 

Take typescript and a ts-based tooling package for exapmle, only when my packages starts using the new features in ts 5.6 increase the peer dependency to 5.6. I'm never supposed to increase it _only because a new version is out_.

jvince:
@SevenOutman 
[...]

PR #971, opened 2024-08-08
feat: add --no-bump-peer-deps flag to prevent bumping peerDependencies

SevenOutman:


src/types/BeachballOptions.ts
```diff
@@ -30,6 +30,7 @@ export interface CliOptions
   all: boolean;
   authType: AuthType;
   bump: boolean;
+  bumpPeerDeps?: boolean;
```
ecraig12345, line 33: Can you move this to the `Pick<RepoOptions, ...>` list above?

ecraig12345:
This looks good. Did you mean to create it as a draft?

SevenOutman:
> This looks good. Did you mean to create it as a draft?

Hi @ecraig12345 thanks for reviewing. This PR was created to fulfill my feature request in #970. I put it in draft status because I wasn't sure if my change works as expected to both add a CLI flag and a configuration field - I myself will be using the configuration.
Could you give me a hint?

src/types/BeachballOptions.ts
SevenOutman: Sure moved.

PR #972, opened 2024-08-21
feat: Enable override to NPM_CONCURRENCY defaults through env variable

nickymunga:
**Description:** 
This update introduces the ability to specify the NPM_CONCURRENCY environment variable, which controls the concurrency level when listing package versions from NPM. The current default settings are either 2 or 5.

**Usage:**
To adjust the concurrency, set the NPM_CONCURRENCY environment variable to the desired value. This change enables users and CI/CD pipelines to override the default concurrency levels when running the beachball sync command.

**Why This Is Necessary:**
For large monorepos, the beachball sync process can take an excessive amount of time due to the high volume of packages. By allowing a higher concurrency setting, this update can help improve performance and reduce the overall sync time.

**Considerations when overriding this value:**

- Network Latency
- Registry Throttling: When interacting with private or public npm registries, excessive concurrency may trigger rate limits.
- Local Resources: High concurrency can increase CPU and memory usage. 

We do not provide a specific recommendation for the ideal NPM_CONCURRENCY value, as this will vary based on your network and system conditions and registry rate limits. We suggest conducting benchmarks to determine the optimal setting for your environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,10 @@ show help message
 
 skips the prompts for publish
 
+## Overriding Concurrency
+
+In large monorepos, the Beachball sync process can be time-consuming due to the high number of packages. To optimize performance, you can override the default concurrency (typically set to 2 or 5) by setting the NPM_CONCURRENCY environment variable to a value that best suits your needs
+
 ## Examples
 
 ```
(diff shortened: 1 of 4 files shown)


Bryan-Cee (approved):
LGTM

ecraig12345 (approved):

Issue #978, opened 2024-09-05
Consider alternative to append-only CHANGELOG.md files


derrickstolee:
I was recently investigating Git repository size bloat and came across Beachball’s CHANGELOG.md files as contributing significantly to repository growth in some internal Javascript repositories. Inefficiencies in Git and Azure DevOps are more at fault here, but those tools are harder to fix on a short timeline. I’m creating this issue to communicate the issue to the Beachball team and see what options they believe are possible to be done here.

For some of the technical Git details, I will be using terms from the following articles:

- [Commits are snapshots, not diffs ](https://github.blog/open-source/git/commits-are-snapshots-not-diffs/)
- [Git's database internals I: packed object store](https://github.blog/open-source/git/gits-database-internals-i-packed-object-store/)

# The Size Issue
I began looking at an internal Javascript monorepo and noticed that the size of the Git blob objects were growing much faster than what would be justified by the number of contributors to the repo. The tree objects were growing at a reasonable rate. This allowed me to focus on the size of the blobs.

As someone familiar with the Git codebase, I was able to prototype a tool that scans the Git object graph grouping object ids by a path they appear at from at least one commit. From these per-path batches, I was able to collect the top 100 paths that contributed to the on-disk size. In this case, I was looking at the packfile that was cloned from Azure DevOps.

Of these top 100 paths, almost all of them were the CHANGELOG.md and CHANGELOG.json files produced by beachball! The only other file in the list was the yarn.lock file that is changed frequently as part of natural development process.

In the worst case, **one CHANGELOG.md file was contributing 1.9 GB to the size of the clone!** This is across 6,769 versions of the file.

This struck me as odd, since it was clear that these CHANGELOG.md files were being appended to and Git should be storing these files using delta compression. Please see the following figure for how these append-only files are stored as full snapshots and then should be stored using efficient delta compression.

![Append-only files in Git](https://github.com/user-attachments/assets/3319b379-e0cb-49e3-b8ad-3684cd1c6e73)

Note how the small gray rectangles do not look like much data. In a well-packed repository, this is indeed how Git would choose to compress these files for efficient storage and network communication.

However, there are two issues that are affecting this process:

1. `git push` is not considering the previous version of the CHANGELOG.md file as a potential delta base, and thus is creating an inefficient packfile to send to the server.
2. Azure DevOps trusts this delta choice and does not recompute the delta base.

We are pursuing fixes to both of these issues, as this inefficiency is affecting more repositories than just the ones using beachball.

[...]

PR #996, opened 2024-10-04
Add option to use hash in changelog filenames

ecraig12345:
Add an option `changelog.uniqueFilenames` which adds an 8-character ~~random~~ suffix to the end of the changelog filename (before the extension): e.g. `CHANGELOG-abcd1234.md`.

EDIT: After discussion, the suffix is now the first 8 characters of the MD5 hash digest of the package name.

This will stay stable between commits, and any existing changelog files will be renamed when modified. Existing suffixed files should also be renamed if the package is renamed. (I'm guessing 8 characters is good enough, but we could also use 12 or more.)

This is one option for working around an issue with Git: its default hash algorithm only considers the last 16 characters of filenames, which can lead to collisions and inefficient packing when many files have similar names, which then leads to extreme growth in size for larger repos with many colliding changelog files.

Bonus: while adding the hash, I learned we can use `crypto.randomUUID()` for change file IDs and remove the `uuid` dependency.

(I also updated a couple things with test configuration: disable type checking in ts-jest since it's redundant and slow, and set launch.json to use Node 14 like the rest of the repo.)

Related to #978

### DIFF
--- a/src/changelog/prepareChangelogPaths.ts
+++ b/src/changelog/prepareChangelogPaths.ts
@@ -8,12 +8,12 @@ interface ChangelogPaths {
   md?: string;
   /** Absolute path to changelog json file (new or existing) */
   json?: string;
-  /** Hash being used in files, if any */
-  hash?: string;
+  /** Suffix being used in files, if any */
+  suffix?: string;
 }
 
 /**
- * Get the paths to the changelog files. Also handles conversion between `changelog.hashFilenames`
+ * Get the paths to the changelog files. Also handles conversion between `changelog.uniqueFilenames`
  * being true and false/unset (moving the files if needed).
  *
  * @returns object with each changelog path, or undefined if that changelog shouldn't be written.
@@ -25,10 +25,10 @@ export function prepareChangelogPaths(params: {
 }): ChangelogPaths {
   const { options, changelogAbsDir: cwd } = params;
   const { changelog, generateChangelog } = options;
-  const { hashFilenames } = changelog || {};
+  const { uniqueFilenames } = changelog || {};
 
   const paths: ChangelogPaths = {};
-  let existingHashedPaths: ChangelogPaths | undefined;
+  let existingSuffixedPaths: ChangelogPaths | undefined;
 
   if (!generateChangelog) {
     return paths;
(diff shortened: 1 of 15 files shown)


src/changelog/prepareChangelogPaths.ts
```diff
@@ -0,0 +1,142 @@
...
+ * Get paths to existing `CHANGELOG-{suffix}.md` and `CHANGELOG-{suffix}.json` files.
+ * Exported for testing only.
+ */
+export function _getExistingSuffixedChangelogs(cwd: string): ChangelogPaths {
+  let suffixedFiles: { file: string; suffix: string; date?: number }[];
+  try {
+    suffixedFiles = fs
+      .readdirSync(cwd)
+      .filter(file => /^CHANGELOG-[a-f\d]{8,}\.(json|md)$/.test(file))
+      .map(file => ({ file, suffix: file.match(/[a-f\d]{8,}/)![0] }));
```
derrickstolee, line 86: This regex really depends on the fact that CHANGELOG- and ".json" or ".md" have enough characters that do not match the hexadecimal representation. To me, it seems like we should strip the file extension and take the last eight characters after that instead of relying on regex magic. (But I have a bias against regexes when string manipulation is straightforward.)

src/changelog/prepareChangelogPaths.ts
```diff
@@ -0,0 +1,142 @@
...
+      continue; // not generating this file type
+    }
+
+    const defaultPath = path.join(cwd, `CHANGELOG.${ext}`);
+
+    if (uniqueFilenames) {
+      // Filenames should use a suffix. Start by getting any existing suffixed changelogs...
+      existingSuffixedPaths ??= _getExistingSuffixedChangelogs(cwd);
+      // Use the existing suffix or generate a new one
+      paths.suffix ??= existingSuffixedPaths.suffix ?? uuid().slice(0, 8);
```
derrickstolee, line 48: This use of a "random" suffix is perhaps concerning to me. The method prototype doesn't currently offer anything else, but I was thinking that having a deterministic suffix based on the directory for the CHANGELOG files (relative to src root) or the package name would be helpful.

The reason to have something like that is to avoid possible collisions if two changelogs are generated in parallel somehow and no longer conflict on merge. Or if there is somehow a bug in the discovery of existing suffixed files, the deterministic algorithm would produce the same filename as the previous time.

It's worth being a bit of a stickler here, since the downside is quite real: if these suffixes start being added randomly, then we are worse off than before in terms of Git finding good deltas. I'd rate this as "low probability to be wrong, but high impact if wrong".

derrickstolee (approved):
Thanks for working on this. I'm not an expert in this codebase or Javascript, so take all of my comments with a grain of salt. I tried to focus on the highest level of functionality, which I believe your implementation will suffice. I prod in a couple of directions for your consideration, but do not consider them blockers.

src/changelog/prepareChangelogPaths.ts
ecraig12345: > This regex really depends on the fact that CHANGELOG- and ".json" or ".md" have enough characters that do not match the hexadecimal representation. 

Not sure what you mean by this? The regex is very specific: `^` start of string, `CHANGELOG-`, 8+ hex characters, `.md` or `.json`, `$` end of string.

derrickstolee: Perhaps I'm misunderstanding things, but I see two regexes and the first filters the filenames based on the match that includes `^` and `$` but the second has a group match that doesn't have those protections. Is this `file.match()` somehow reusing the previous regex information?

So what I was trying to say is that the regex `/[a-f\d]{8,}/` will find a match, and relies on the fact that a string matching `/^CHANGELOG-[a-f\d]{8,}\.(json|md)$/` has only one possible substring to match that hex string. (This would not be the case for a string `deadbeef-12345678.json`)

ecraig12345: Sorry about that, for some reason I initially read your comment as pointing to the next line up (filter). Then after posting I realized you were actually talking about the map line, which makes more sense.

src/changelog/prepareChangelogPaths.ts
ecraig12345: Thanks, I was sort of wondering about the random suffix vs. hashing something. I think if we're actually hashing something, it should be the package name, since that's probably more stable than the path. I'll just need to find a different library for hashing (`uuid` is only for random GUIDs).

src/changelog/prepareChangelogPaths.ts
```diff
@@ -41,30 +39,26 @@ export function prepareChangelogPaths(params: {
...
-      // Filenames should not use a suffix
-      paths[ext] = defaultPath;
-
-      if (!fs.existsSync(defaultPath)) {
-        // If the default file doesn't exist, check for a suffixed file to move back
-        existingSuffixedPaths ??= _getExistingSuffixedChangelogs(cwd);
-        if (existingSuffixedPaths[ext]) {
-          moveIfNeeded({ oldPath: existingSuffixedPaths[ext]!, newPath: defaultPath, hadSuffix: true });
+    // Generate a unique filename based on the package name hash.
+    const hash = crypto.createHash('md5').update(packageName).digest('hex').slice(0, 8);
```
derrickstolee, line 43: Thank you for this change. This will avoid the possibility of some corner cases with the random method. There may even be some benefits if a package is moved within the repo (without renaming the package).

derrickstolee (approved):

src/changelog/prepareChangelogPaths.ts
```diff
@@ -0,0 +1,109 @@
...
+ * Get the path to the newest `CHANGELOG-{hash}.{ext}` file in the given directory.
+ */
+function getExistingHashedPath(params: { cwd: string; ext: 'md' | 'json' }): string | undefined {
+  const { cwd, ext } = params;
+  const fileRegexp = new RegExp(`^CHANGELOG-[a-f\\d]{8}\\.${ext}$`);
+  try {
+    let newestDate = 0;
+    let newestFile: string | undefined;
+
+    for (const file of fs.readdirSync(cwd)) {
```
dmichon-msft, line 80: Best practice would be to pass the `{ withFileTypes: true }` option and test that the entry is a file.

src/changelog/prepareChangelogPaths.ts
```diff
@@ -0,0 +1,109 @@
...
+  } catch (e) {
+    console.warn(`Error getting changelog file info in ${cwd}: ${e}`);
+  }
+}
+
+function renameIfPresent(params: { oldPath: string; newPath: string }): boolean {
+  const { oldPath, newPath } = params;
+  try {
+    if (fs.existsSync(oldPath)) {
+      fs.renameSync(oldPath, newPath);
```
dmichon-msft, line 101: Existence checking files before operating on them is generally an anti-pattern that is advised against by the authors of the file system APIs, since in theory the existence can change between the calls, so you have to handle the non-existence during the real call anyway. Simpler to just handle the `ENOENT` or `ENOTDIR` during the `renameSync` call.

PR #1026, opened 2024-12-09
docs: fix grammar mistake

muningis:
Documentation on https://microsoft.github.io/beachball/overview/configuration.html page says ```
/** @type {import('beachball').BeachallConfig} */
```
instead of
```/** @type {import('beachball').BeachballConfig} */```


ecraig12345 (approved):
Thanks!

PR #1042, opened 2025-03-31
updateRelatedChangeTypes: fix cubic behavior

derrickstolee:
The `updateRelatedChangeType()` method can be called with individual packages or a list of packages. In either case, this presents a _minimum_ of quadratic performance as it does a breadth-first-search of the dependency graph starting at each input package, one a at a time. For example, in an internal monorepo that recently had a Typescript upgrade initiate 3000+ package updates at once, `bumpInPlace()` calls this method 3000+ times and each does a bunch of repeated work.

To resolve this issue, we will need to combine things into a single DAG walk. That is a more involved change that will take more time and testing. I will report back later. We need something to resolve the pain sooner.

The current implementation has a more immediate issue: the quadratic behavior is actually sometimes _cubic_. The point here is that there are two data structures handling the BFS walk:

1. `queue` is a list that handles the boundary of the walk. These are the nodes we want to explore next.
2. `visited` is a set that prevents exploring a node multiple times.

The issue here is that the `visited` set needs to protect insertion into `queue`, not just prevent repeat visiting a node as we pop off of the queue. As it stands, we will insert a dependent package into the queue for each package that depends on it. This adds space and time concerns that contribute to the slowness of this method.

This PR moves the use of the `visited` set interactions into the locations that care about pushing into `queue`, lowering the space and time constraints around the queue.

### DIFF
--- a/src/bump/updateRelatedChangeType.ts
+++ b/src/bump/updateRelatedChangeType.ts
@@ -98,11 +98,7 @@ export function updateRelatedChangeType(params: {
 
       if (group) {
         for (const packageNameInGroup of group.packageNames) {
-          if (!group.disallowedChangeTypes?.includes(updatedChangeType)) {
-            if (visited.has(packageNameInGroup)) {
-              continue;
-            }
-
+          if (!group.disallowedChangeTypes?.includes(updatedChangeType) && !visited.has(packageNameInGroup)) {
             visited.add(packageNameInGroup);
             queue.push({
               subjectPackage: packageNameInGroup,
--- a/src/bump/updateRelatedChangeType.ts
+++ b/src/bump/updateRelatedChangeType.ts
@@ -53,12 +53,6 @@ export function updateRelatedChangeType(params: {
     while (queue.length > 0) {
       const { subjectPackage, changeType } = queue.shift()!;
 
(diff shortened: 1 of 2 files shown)


derrickstolee:
@ecraig12345 Could you please take a look at this and let me know what you think?

derrickstolee:
Latest push includes a change that might be helpful: it avoids propagating the necessary change type if we haven't changed the type along the way. This will reduce how many nodes of the dependency DAG we need to visit each time, especially when the loop in `bumpInPlace()` iterates on multiple packages.

src/bump/updateRelatedChangeType.ts
```diff
@@ -92,6 +99,11 @@ export function updateRelatedChangeType(params: {
       if (group) {
         for (const packageNameInGroup of group.packageNames) {
           if (!group.disallowedChangeTypes?.includes(updatedChangeType)) {
+            if (visited.has(packageNameInGroup)) {
+              continue;
+            }
+
```
ecraig12345, line 105:
```suggestion
          if (!group.disallowedChangeTypes?.includes(updatedChangeType) && !visited.has(packageNameInGroup)) {
```

ecraig12345 (approved):

PR #1079, opened 2025-10-24
feat: Introduce --fixup to change subcommand

evancharlton:
When working on a change in a monorepo, I will often need to add subsequent change files as the change spreads into multiple packages. This creates multiple changefiles as well as multiple commits.

To tackle this, this change introduces a `--fixup` flag, modeled after the `--fixup` flag in `git commit`. When used, this will enable fix-up mode, which will:

 1. Find the most recent commit which adds a change file
 2. Add the new changes into that file
 3. Commit with a `--fixup <sha>` flag

If no commits are found in the current branch, then the standard flow is used to create a new commit / changefile.

### DIFF
--- a/change/beachball-d725cdf6-20dd-41df-9402-cf9cf4f9bc08.json
+++ b/change/beachball-d725cdf6-20dd-41df-9402-cf9cf4f9bc08.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: Introduce --fixup to change subcommand",
+  "packageName": "beachball",
+  "email": "evancharlton@microsoft.com",
+  "dependentChangeType": "patch"
+}


src/changefile/writeChangeFilesFixup.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/**
+ * Finds the fork point (merge base) with the main branch.
+ * @param cwd - Working directory
+ * @returns The commit hash of the fork point, or null if not found
+ */
+async function findForkPoint(cwd: string): Promise<string | null> {
+  // Try common main branch names
+  const mainBranches = ['origin/main', 'origin/master', 'main', 'master'];
```
ecraig12345: This should use `BeachballOptions.branch` as the comparison basis

src/changefile/writeChangeFilesFixup.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  const commits = logResult.stdout
+    .trim()
+    .split('\n')
+    .filter(line => line.trim());
+
+  // Search through commits from newest to oldest
+  for (const commitHash of commits) {
+    // Check what files were added in this commit (compare with parent)
+    const diffResult = await gitAsync(
```
ecraig12345: There's already some closely related logic in `getChangedPackages`, so it would be best if at least some of that could be reused/combined with the logic here.

src/changefile/writeChangeFilesFixup.ts
```diff
@@ -0,0 +1,242 @@
...
+        const normalizedChangePath = relativeChangePath.replace(/\\/g, '/');
+
+        if (normalizedPath.startsWith(normalizedChangePath) && normalizedPath.endsWith('.json')) {
+          const fileName = path.basename(normalizedPath);
+          addedChangeFiles.push(fileName);
+        }
+      }
+    }
+
+    if (addedChangeFiles.length > 0) {
```
ecraig12345: It seems like this will only work properly with grouped change files? For ungrouped change files, you either need to find the right file for the specific package, or just create a new file and add it to the commit with change files. Creating a new file is most similar to what you're doing with grouped changes.

src/changefile/writeChangeFilesFixup.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  // Update the change file with new changes
+  if (groupChanges) {
+    // For grouped changes, add to the existing changes array
+    const groupedChanges = existingChangeInfo as { changes: ChangeFileInfo[] };
+    groupedChanges.changes.push(...changes);
+    fs.writeFileSync(changeFilePath, JSON.stringify(groupedChanges, null, 2));
+  } else {
+    // For individual change files, merge the new change with the existing one
+    // We assume there's only one new change for fixup mode with individual files
```
ecraig12345: I think in this case it's better to just add new change files rather than modifying an existing one--that's the equivalent of what you're doing in grouped mode. Or if the intent was to modify the existing change for a package (if present), you'd need some updates for both the grouped and ungrouped cases to handle that consistently. 

Also I think the current logic is overwriting whichever package's change file was found first, which might not be the correct one for that package. (Most multi-package repos use grouped change files, but there are a few that don't.)

ecraig12345 (changes requested):
This is a good idea but needs a few updates for behavior consistency between grouped/ungrouped and avoiding duplicated logic.

PR #1080, opened 2025-11-07
Fix bump logic for deps using file protocol

### LINKED ISSUE #1035: Local packaged dependencies referenced by file mistakenly trigger bump and show up in Changelog
When a package has internal private dependencies (i.e. packaged utils referenced by "file:" in the package.json), the existing version range will not match the package's actual version. This is causing beachball bump to add changelog entries on bump referencing the bumping of said internal dependencies.

https://github.com/microsoft/beachball/blob/3f71c577870d1cedea81495888950e47179d4a7a/src/bump/setDependentVersions.ts#L36

I couldn't find any configuration option in the docs to make beachball ignore these internal packages.

Example from a verbose execution of `beachball bump`:

```bash
@azure/msal-common needs to be bumped because eslint-config-msal file:../../shared-configs/eslint-config-msal -> 0.0.0
@azure/msal-common needs to be bumped because msal-test-utils file:../../shared-test-utils -> 0.0.1
@azure/msal-common needs to be bumped because rollup-msal file:../../shared-configs/rollup-msal -> 0.0.0
```

Example of the resulting changelog with false bump entries for these dependencies:

https://github.com/AzureAD/microsoft-authentication-library-for-js/commit/20674a197d20fb865888d7771f5edd2dc974e854#diff-8f1fef2ac7777148212f5d8a43e43e428ba981830435fcef371908663229d89eR14

Is there anything in the config that can currently prevent this?

tnorling:
We have several devDependencies defined using the file protocol e.g. `"internal-package": "file:../../internalPackage"` and beachball includes these packages in our changelog every time we bump because the dependency version defined doesn't match the version `0.0.0` defined in the internal package's package.json.

This PR applies the same logic to this as setting the version to '*'. I'm not sure if there's other areas that need to be fixed as well but this seems to address the problem we're having.

Fixes #1035

### DIFF
--- a/change/beachball-25e2af71-4356-43fb-9a17-7762f74724d6.json
+++ b/change/beachball-25e2af71-4356-43fb-9a17-7762f74724d6.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "Update dependency workspace-tools to ^0.40.0. This may improve detection of remote branch names in certain cases. It also includes stricter error checking for git operations, which may surface some errors which were previously hidden.",
+  "packageName": "beachball",
+  "email": "elcraig@microsoft.com",
+  "dependentChangeType": "patch"
+}
--- a/change/beachball-3c271966-bbc3-44a7-8c90-e4e71c02aaa5.json
+++ b/change/beachball-3c271966-bbc3-44a7-8c90-e4e71c02aaa5.json
@@ -1,0 +1,7 @@
+{
+  "comment": "Refresh lock again",
+  "type": "none",
+  "packageName": "beachball",
+  "email": "elcraig@microsoft.com",
+  "dependentChangeType": "none"
+}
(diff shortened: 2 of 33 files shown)


ecraig12345:
This makes sense, but can you add a test for this scenario in `src/__e2e__/bump.test.ts` to make sure nothing breaks elsewhere?

Also just FYI in case you weren't aware, with yarn v2+ and pnpm, you can use [`workspace:` versions](https://yarnpkg.com/features/workspaces#cross-references) to solve this same problem. Though that doesn't help if you're on yarn v1 (not sure about npm).

tnorling:
> This makes sense, but can you add a test for this scenario in `src/__e2e__/bump.test.ts` to make sure nothing breaks elsewhere?
> 
> Also just FYI in case you weren't aware, with yarn v2+ and pnpm, you can use [`workspace:` versions](https://yarnpkg.com/features/workspaces#cross-references) to solve this same problem. Though that doesn't help if you're on yarn v1 (not sure about npm).

Test added. I did see that while investigating this but we use npm where the workspace protocol isn't supported in dependencies

ecraig12345:
Ah, I see what you mean now. There's no regression from the change, I just misunderstood the original issue you were trying to address.

PR #1089, opened 2025-11-20
Dependency validation and formatting improvements

ecraig12345:
Validate that there are no private packages among `peerDependencies` or `optionalDependencies` of published packages (not just `dependencies`). There's a slight chance this could be a breaking change.

Copy the `bulletedList` helper from Cloudpack and update list formatting in CLI output. If someone is parsing `beachball check` or `beachball publish` output for some reason, this could be a breaking change.

### DIFF
--- a/change/beachball-0ae4a3b0-1089-4a2a-a577-c0f9b4c550b2.json
+++ b/change/beachball-0ae4a3b0-1089-4a2a-a577-c0f9b4c550b2.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Validate that there are no private packages among `peerDependencies` or `optionalDependencies` of published packages (not just `dependencies`). There's a slight chance this could be a breaking change.",
+  "packageName": "beachball",
+  "email": "elcraig@microsoft.com",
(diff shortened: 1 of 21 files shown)


timmevandermeer:
this indeed created a issue for my use case. I have private workspace dependencies that are still marked as peer dependencies to make sure it is not bundled and shared with another workspace package.

The structure is basically package A (container app, published) -> package B (module, published but only for internal consumption through a manifest) -> package C (a common singleton lib, dep of A and peer dep of B, private in repo). 

Anyway, this is a bit of an esoteric use case but the change does block me from upgrading right now.

ecraig12345:
@timmevandermeer Interesting, wouldn't that cause errors when consumers try to install your package, since a referenced peer dep doesn't exist?

timmevandermeer:
@ecraig12345 well that's why I say it's a niche use case. These in my example B is essentially a private workspace package that gets published to a private registry for the sole reason that package A in some cases is deployed with a specific version of B instead of the workspace version. Perhaps the solution is to simply publish C. But I feel like beachball could benefit from a bit more configuration options to make these check less blocking

PR #1206, opened 2026-04-18
Security: NPM auth token is passed on command line (process list exposure)

tomaioo:
## Summary

Security: NPM auth token is passed on command line (process list exposure)

## Problem

**Severity**: `High` | **File**: `packages/beachball/src/packageManager/npmArgs.ts:L20`

Publish arguments include auth credentials directly as CLI flags (for example `--//registry/:_authToken=<token>`). Command-line arguments can be exposed via process listings, CI diagnostics, crash reports, or audit tooling, leading to credential leakage.

## Solution

Do not pass tokens in CLI args. Prefer environment variables supported by npm (`NODE_AUTH_TOKEN`) or write a temporary `.npmrc` with restrictive permissions and delete it after use. Ensure logs never print full auth material.

## Changes

- `packages/beachball/src/packageManager/npmArgs.ts` (modified)

### DIFF
--- a/packages/beachball/src/packageManager/npmArgs.ts
+++ b/packages/beachball/src/packageManager/npmArgs.ts
@@ -11,6 +11,10 @@ export function getNpmLogLevelArgs(verbose: boolean | undefined): string[] {
 export function getNpmPublishArgs(packageInfo: PackageInfo, options: Omit<NpmOptions, 'path'>): string[] {
   const { registry, access } = options;
   const authArgs = getNpmAuthArgs(options);
+  if (authArgs) {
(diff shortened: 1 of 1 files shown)


ecraig12345 (changes requested):
This is something that could potentially be handled better, but npm itself doesn't support `NODE_AUTH_TOKEN`--best I can tell it's [specific to `actions/setup-node`](https://docs.github.com/en/actions/tutorials/publish-packages/publish-nodejs-packages), and if you follow those instructions you could already pass the token that way without modifying beachball.

ecraig12345:
Closing in favor of #1211

PR #1211, opened 2026-04-21
Handle npm tokens with environment variables

ecraig12345:
Add support for reading the npm token from the `NPM_TOKEN` environment variable, and update docs to recommend that approach. Getting the token from the environment is handled in `getCliOptions` for simplicity.

Pass the token through to `npm` using an environment variable rather than a CLI option. It's not well-documented for the specific case of auth, but the environment variable naming format is `npm_config_//registry.npmjs.org/:_authToken` (or other registry; trailing slash on registry is required).

### Workaround for token env issues

Special characters such as `/` and/or `:` aren't valid in env var names for certain shells/platforms, which caused the token environment variable to be dropped when trying to pass it to the `npm publish` subprocess. Unclear whether the issue is specific to platform, shell, or the combo, but it definitely happened with Ubuntu+bash.

First issue is that running `npm publish` with `shell: true` drops the token var. Fixed by only setting `shell: true` on Windows.

The other issue is specific to if the parent command was run VIA `yarn`... Best I can tell, this is because on every `yarn run ...` command, yarn makes temp directories like `/<temp>/yarn--1776822418161-0.7992675923334178` with aliases for `node` and `yarn`. On Linux (and Mac), the `node` alias looks something like:
```sh
#!/bin/sh
exec "/path/to/node" "$@"
```
My best guess is this extra `exec` step drops invalid environment variable names?? (This consistently reproed on Ubuntu+bash, but not Mac+zsh or bash. The clue was that the tests passed even on Linux when run via debugTests.js, but failed when run via yarn test.) Removing the `yarn--` segment from the PATH seems to consistently fix this issue.


packages/beachball/src/__fixtures__/registry.ts
```diff
@@ -99,8 +106,54 @@ export class Registry {
...
+    if (!this.isLoggedIn) {
+      throw new Error('Must be logged in to get the token');
+    }
+
+    const registryUrl = this.getUrl().replace(/^https?:/, '');
+    const npmrcPath = path.join(os.homedir(), '.npmrc');
+    const npmrcContent = fs
+      .readFileSync(npmrcPath, 'utf-8')
+      .split(/\r?\n/g)
+      .find(line => line.startsWith(registryUrl) && line.includes('_authToken'));
```
dannyvv, line 141: nit: it could also be `_password`

dannyvv (approved):

packages/beachball/src/__fixtures__/registry.ts
ecraig12345: In these tests using `npm login` against a Verdaccio registry, it always seems to use `_authToken`, so this should be fine.

PR #1220, opened 2026-04-28
Include catalog dependency changes when checking for changed packages

ecraig12345:
## Issue

Supposing a repo defines `catalog`/`catalogs` in `.yarnrc.yml`, and then it changes in a branch, e.g.:
```diff
  catalog:
-   react: ^18.0.0
+   react: ^19.0.0
```
The catalog dependency in a package looks like this in `package.json`, and by design, the file itself doesn't change:
```json
{
  "name": "@scope/foo",
  "dependencies": {
    "react": "catalog:"
  }
}
```
With the initial implementation of catalog support in beachball, catalog-dependent packages like `@scope/foo` would not be listed as changed when the dependency's catalog version changes.

## Fix

This PR updates the `getChangedPackages` logic to use a new function `getCatalogChangedPackages` which checks for catalog diffs:
- If the repo defines catalogs *and* the file with the catalog definition has changed, it gets the catalog file content from the target branch and compares the catalog objects. 
- If there are any changes, it checks all included packages to get the list of packages using those catalog dependencies.

Most of the other changes are refactoring to support passing the previously-retrieved list of changed files to `getCatalogChangedPackages`, so we can skip the extra git operation of retrieving the file from the target branch if we already know it didn't change (which is by far the most common case). Removing one git operation doesn't matter much in small repos, but becomes important in extremely large repos.


packages/beachball/src/__functional__/git/ensureSharedHistory.test.ts
```diff
@@ -18,15 +18,15 @@ describe('ensureSharedHistory', () => {
   let repositoryFactory: RepositoryFactory;
   const logs = initMockLogs();
   const testBranch = 'test';
+  const defaultRefSpec = `+refs/heads/${defaultBranchName}:refs/remotes/${defaultRemoteBranchName}`;
```
ecraig12345, line 21: Ignore this file

packages/beachball/src/__functional__/git/fetch.test.ts
```diff
@@ -27,13 +27,18 @@ describe('gitFetch', () => {
   /** To speed things up, some tests only check the arguments and skip the git operation */
   const noOpSuccess = () => ({ success: true, stdout: '', stderr: '', status: 0 } as GitProcessOutput);
 
-  const realGit = jest.requireActual<typeof workspaceTools>('workspace-tools').git;
+  const { git: realGit, gitFailFast: _realGitFailFast } = jest.requireActual<typeof workspaceTools>('workspace-tools');
```
ecraig12345, line 30: Ignore this file

packages/beachball/src/__tests__/publish/__snapshots__/bumpAndPush.test.ts.snap
```diff
@@ -5,15 +5,15 @@ exports[`bumpAndPush retries on fetch failure then succeeds 2`] = `
 [log] Bumping versions and pushing to git (attempt 1/5)
 [log] Reverting
 [log]
-[log] Fetching branch "master" from remote "origin"...
+[log] Fetching branch "master" from remote "origin" (+refs/heads/master:refs/remotes/origin/master)...
```
ecraig12345, line 8: Ignore this file and corresponding test

packages/beachball/src/git/ensureSharedHistory.ts
```diff
@@ -6,8 +6,14 @@ import { BeachballError } from '../types/BeachballError';
 
 /**
  * Ensure that adequate history is available to check for changes between HEAD and `options.branch`.
- * Otherwise attempting to get changes will fail with an error "no merge base".
- * (This is mostly an issue with CI builds that use shallow clones.)
+ * Otherwise attempting to get changes will fail with an error "no merge base". (This is mostly an
```
ecraig12345, line 9: Ignore this file

packages/beachball/src/git/fetch.ts
```diff
@@ -3,10 +3,16 @@ import { getGitEnv } from './gitAsync';
 
 type GitFetchParams = {
   cwd: string;
-  /** Remote to fetch from. If not specified, fetches all remotes. */
-  remote?: string;
-  /** Branch to fetch. This will be ignored if `remote` is not also specified. */
-  branch?: string;
+  /**
```
ecraig12345, line 6: Ignore this file

packages/beachball/src/changefile/getCatalogChangedPackages.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  const changedPackages: string[] = [];
+
+  for (const pkg of Object.values(packageInfos)) {
+    // Skip logging exclude reasons for catalog references
+    if (!isPackageIncluded(pkg, scopedPackages).isIncluded) continue;
+
+    const changedDeps: string[] = [];
+
+    for (const depType of consideredDependencies) {
```
christiango, line 77: Minor nit, would it be more efficient to instead of looping over all deps for the package, iterate over all the changed catalog packages and check if the package has a catalog reference to that dep?

christiango (approved):

packages/beachball/src/changefile/getCatalogChangedPackages.ts
ecraig12345: Good point, I think I could flatten the names of maybe-catalog deps and only do detailed checks for those keys.

PR #1359, opened 2026-07-29
Add support for passing git token via env

ecraig12345:
## Summary

Adds a `gitToken` option (read from the `BEACHBALL_GIT_TOKEN` environment variable) so Beachball can authenticate git fetch/push with minimal risk of exposing the token. Also adds `BEACHBALL_NPM_TOKEN` as an alternative to `NPM_TOKEN` for npm auth.

## Motivation

Previously, passing a git token required manually setting it in the remote URL or `http.<url>.extraheader` config, which risks the token leaking into logs or in other ways. This PR provides first-class support for a safer path, including handling for overriding existing `http.<url>.extraheader` configs automatically set by the CI checkout step.

## How it works

- `getGitAuthEnv` builds env-based git config (`GIT_CONFIG_*`) that injects an `Authorization: Basic <base64 x-access-token:token>` header to spawned git processes. Because it's passed via env, the token never appears on the command line or in `.git/config`.
- The header is scoped to the resolved remote URL (`http.<remoteUrl>.extraheader`), so it isn't sent to cross-host redirect targets.
- Existing `extraheader` config that applies to the remote is rebuilt: any **auth** headers (e.g. a token injected by a CI checkout step) are **dropped**, while **non-auth** headers are **preserved** (re-added under the remote-URL key so they survive git's specificity-ordered reset).
- Git trace/curl-verbose env vars are forced off so the header can't leak via tracing.
- Applied to both the fetch path (`ensureSharedHistory` / `gitFetch`) and the push path (`bumpAndPush`).
- `gitToken` is env-only (excluded from CLI option definitions).

[More details about git `extraheader` overrides and other approaches considered](https://gist.github.com/ecraig12345/db3af8943e697eff3a847d954280227c)

The name `BEACHBALL_GIT_TOKEN` was chosen to avoid interference with git's semi-standard `GIT_*` env vars, or anything else which might use a generic `GIT_TOKEN` env for some other purpose.

## Changes

- **New:** `src/git/getGitAuthEnv.ts` — builds the scoped auth env, with header encode/decode helpers.
- **Modified:** all git fetch and push operations pass `gitToken` through and apply the auth env.
- **Modified:** add `gitToken` to `CliOptions` for convenience (though it's not accepted by the CLI) and read `BEACHBALL_GIT_TOKEN` from env
  - Also add `BEACHBALL_NPM_TOKEN` as a consistent alternative for `NPM_TOKEN`
- **CI:** `scripts/publishWithGithubAppToken.ts`, `.ado/release.yml`, `.github/workflows/release.yml` — pass the token via `BEACHBALL_GIT_TOKEN`; removed the wrapper's manual `extraheader` manipulation now handled natively.
- **Docs:** Update config docs and migration guide

## Tests

- `src/__tests__/git/getGitAuthEnv.test.ts` — unit test
- `src/__functional__/git/getGitAuthEnvReal.test.ts` — pushes to a real in-process HTTP remote to prove the override approach works as expected at runtime with `git`


packages/beachball/src/git/getGitAuthEnv.ts
```diff
@@ -0,0 +1,218 @@
...
+  //
+  // These must be re-added under the *remote URL* key (not their original keys): git accumulates
+  // `extraHeader` values by specificity and an empty value resets the whole accumulated list, so the
+  // empty reset below (at the most-specific remote URL key) wipes any header re-added under a
+  // less-specific key (e.g. the base `http.extraheader`). Re-adding under the remote URL key is the
+  // only placement that survives the reset. This narrows a base header to the remote URL, which is
+  // acceptable (and safer against leaking to redirect targets) for a single-remote fetch/push.
+  const preservedHeaders = getExistingHeaders(cwd)
+    .filter(({ key, value }) => _doesHeaderApply({ key, remoteUrl }) && !/^\s*authorization\s*:/i.test(value))
+    .map(({ value }) => value);
```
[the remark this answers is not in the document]
ecraig12345: Good to flag, but this doesn't apply in beachball's usage, so no change is needed:

- `getGitAuthEnv` is only ever called with `env: process.env` (from `bumpAndPush` and `fetch`).
- `getExistingHeaders` reads existing config via `git config --get-regexp`, which is spawned inheriting `process.env` — and `git config` **does** report env-provided `GIT_CONFIG_*` entries. So an env-provided `http.extraheader=X-Proxy: value` **is** collected, kept (it's non-auth), and re-added under the remote-URL key, meaning it survives the reset rather than being silently lost.

Verified against real git:

```
$ GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=http.extraheader GIT_CONFIG_VALUE_0='X-Proxy: value' \
    git config --get-regexp '.*\.extraheader'
http.extraheader X-Proxy: value
```

The failure mode described here would require a caller to pass an `env` argument that diverges from the actual process environment (carrying extraheaders not present in `process.env`). That never happens in beachball — the param is documented as "`process.env` or an override for testing" and is always `process.env` in production. Happy to add a comment noting that invariant if it's useful.

packages/beachball/src/git/getGitAuthEnv.ts
```diff
@@ -0,0 +1,216 @@
...
+  if (!gitToken) {
+    return undefined;
+  }
+
+  const cacheKey = `${cwd}\0${remote}\0${operation}\0${gitToken}`;
+  let cached = authEnvCache.get(cacheKey);
+  if (!cached) {
+    // Get the actual remote URL to figure out if any existing headers apply, and how to scope
+    // the new auth header (the push URL when building the env for a push).
+    const remoteUrl = getRemoteUrl(params);
```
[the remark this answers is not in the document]
ecraig12345: Not a major concern in practice, and if there were changes made to address this, it should probably be at the workspace-tools level for all the git commands

.github/workflows/release.yml
```diff
@@ -49,16 +49,8 @@ jobs:
...
-
-          # Get the existing remote URL without creds, and use a trap (like try/finally)
-          # to restore it after this step finishes
-          trap "git remote set-url origin '$(git remote get-url origin)'" EXIT
-
-          # Add a token to the remote URL for auth during release
-          git remote set-url origin "https://$REPO_PAT@github.com/$GITHUB_REPOSITORY"
-
+          git config user.email "257645319+office-ogx-auth-helper[bot]@users.noreply.github.com"
+          git config user.name "OGX bot"
```
Saadnajmi, line 53: 👌

docs/concepts/ci-integration.md
```diff
@@ -87,21 +87,26 @@ The [built-in `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides
...
+  - Any CI platform + Azure key vault: [beachball v3's `github-app-token` helper](https://github.com/microsoft/beachball/blob/main/packages/beachball/src/githubAuth/README.md)
   - GitHub Actions + Azure key vault: [`microsoft/create-github-app-token-via-key-vault`](https://github.com/microsoft/create-github-app-token-via-key-vault)
-  - GitHub Actions + GitHub environment secrets: [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token)
+  - GitHub Actions + GitHub environment secrets: [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token), with the private key stored [as a secret](#storing-secrets)
 
-After creating the token, there are various ways it can be passed through to the `git` commands run within `beachball publish`. The most common approach is to set it as the git remote URL. For example, if the token is in an environment variable called `REPO_PAT`, and the remote is called `origin`:
+After creating the token, it can be passed through to `beachball publish` via the `BEACHBALL_GIT_TOKEN` environment variable. You'll also need to set a username and email before publishing. For example:
```
Saadnajmi: nonblocking: This confused me a little, but then GPT Sol suggested something like this:
> `BEACHBALL_GIT_TOKEN` authenticates fetch and push. If Beachball creates a commit, ensure Git has `user.name` and `user.email` configured.

docs/concepts/ci-integration.md
```diff
@@ -87,21 +87,26 @@ The [built-in `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides
...
-  - Any CI platform + Azure key vault: [beachball's `github-app-token` helper](https://github.com/microsoft/beachball/blob/main/packages/beachball/src/githubAuth/README.md)
+- New: use a **GitHub app installation token**. For this purpose, an "app" is essentially just an _identity with permissions_; you don't need to define any logic or endpoints. Create a GitHub app, install it in your repo, and give it permission to bypass policies, then set up one of the following:
+  - Any CI platform + Azure key vault: [beachball v3's `github-app-token` helper](https://github.com/microsoft/beachball/blob/main/packages/beachball/src/githubAuth/README.md)
   - GitHub Actions + Azure key vault: [`microsoft/create-github-app-token-via-key-vault`](https://github.com/microsoft/create-github-app-token-via-key-vault)
-  - GitHub Actions + GitHub environment secrets: [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token)
+  - GitHub Actions + GitHub environment secrets: [`actions/create-github-app-token`](https://github.com/actions/create-github-app-token), with the private key stored [as a secret](#storing-secrets)
```
Saadnajmi, line 93: Should we order the new app based approach first so more users are funneled that way? Or do we expect most people will use fine grained tokens?

docs/concepts/ci-integration.md
```diff
@@ -133,7 +138,7 @@ The exact publishing setup will vary depending on your CI setup, but the overall
    git config user.name "someone"
    git config user.email "someone@example.com"
    ```
-2. Set up git authentication. This could use tokens (covered below), SSH keys, or some other non-interactive method.
+2. Set up git authentication. This could use a token (passed to beachball via `BEACHBALL_GIT_TOKEN`, as covered above), SSH keys, or some other non-interactive method.
```
Saadnajmi: non blocking: Could we have a markdown link to that header? instead of "as covered above", not sure if that reads cleaner since it's the same page

docs/concepts/ci-integration.md
```diff
@@ -194,40 +188,31 @@ Here's a sample setup for publishing from a GitHub repo using Azure Pipelines. T
 This sample assumes the following:
 
 - A variable group called `Beachball secrets` (set up [as described above](#secrets-azure-pipelines)) with the following secrets:
-  - `REPO_PAT`: A GitHub fine-grained personal access token with write access ([as described above](#github-repos))
   - `NPM_TOKEN`: An npm token with write access to the package(s) and/or scope(s), such as a [fine-grained token for public npm](#trusted-publishing-preferred)
+  - `REPO_PAT`: A GitHub fine-grained personal access token with write access ([as described above](#github-repos)).
+    - Alternatively, this could be an app installation token configured in an earlier step rather than stored in the key vault.
```
Saadnajmi: should we rename REPO_PAT to be a little more descriptive so you need less explanation?

scripts/publishWithGithubAppToken.ts
```diff
@@ -1,36 +1,33 @@
...
 //
-// This script sets credentials from the given GITHUB_APP_TOKEN, then runs publish commands.
-// It's meant to be run in ADO pipelines only.
+// This script configures the git identity, runs `beachball publish --no-publish` with BEACHBALL_GIT_TOKEN,
+// and revokes the token on exit. It's meant to be run in ADO pipelines only.
 //
 
 import { spawnSync } from 'child_process';
+import fs from 'fs';
 import path from 'path';
```
Saadnajmi, line 10: nit: node:fs, etc

scripts/publishWithGithubAppToken.ts
```diff
@@ -1,36 +1,33 @@
...
 import { findGitRoot, git, type GitProcessOutput } from 'workspace-tools';
 
 const repoRoot = findGitRoot(process.cwd());
 
-const token = process.env.GITHUB_APP_TOKEN;
+const token = process.env.BEACHBALL_GIT_TOKEN;
 const gitUserEmail = '257645319+office-ogx-auth-helper[bot]@users.noreply.github.com';
 const gitUserName = 'OGX bot';
```
Saadnajmi, line 17: I don't suppose we have some more neutral name than OGX bot, given OGX will get reorged and renamed for sure? I think I just used gitHub-actions[bot] and react-native-sdk[bot]

packages/beachball/src/git/ensureSharedHistory.ts
```diff
@@ -19,9 +19,9 @@ import { BeachballError } from '../types/BeachballError';
  * Throws an error if history is inadequate and cannot be fixed.
  */
 export function ensureSharedHistory(
-  options: Pick<BeachballOptions, 'fetch' | 'path' | 'branch' | 'depth' | 'verbose'>
+  options: Pick<BeachballOptions, 'fetch' | 'path' | 'branch' | 'depth' | 'verbose' | 'gitToken'>
```
Saadnajmi, line 22: non blocking: shoould this be pulled out into it's own type?

Saadnajmi (approved):

docs/concepts/ci-integration.md
ecraig12345: I think it's clearer how the token is used in the full context of the section, but I updated the wording about the name and email.

docs/concepts/ci-integration.md
ecraig12345: I could go either way with this... the PAT approach is simpler, and I still don't see enough benefit to the app approach to strongly suggest people go to the extra trouble if it's not required by organization policies.

docs/concepts/ci-integration.md
ecraig12345: Actually I removed the name and email part from this section since it's explained below.

scripts/publishWithGithubAppToken.ts
ecraig12345: Haha good question... 😅 in this case there's no good generic name since the app is used across multiple repos that don't really have a unifying name/concept (we're not calling it "extended-historic-kenfluentverse-bot") so a random org name that might change is probably fine. And I don't think we should use github-actions[bot] here since that creates confusion that we're using the built-in token (granted continuing to use Ken's name when it's been my token for years is not much better...).

scripts/publishWithGithubAppToken.ts
ecraig12345: This project still pretty much uses the old bare names, so I'll probably wait to fix that until some future all-up change.

packages/beachball/src/git/ensureSharedHistory.ts
ecraig12345: Generally the pattern I've used is that the type only gets pulled out if it's reused somewhere else, which this one isn't.

PR #1362, opened 2026-08-03
Support managed identity based authentication with ESRP

Abe27342:
ESRP's documentation now recommends using an allow-listed managed identity to publish packages rather than an app registration. This adds support for that form of authentication in esrp-npm-release as an alternative to the app registration + certificate approach.

See documentation updates in the README.md for details on how to use this authentication approach.

### DIFF
--- a/packages/esrp-npm-release/README.md
+++ b/packages/esrp-npm-release/README.md
@@ -452,6 +452,10 @@ Append one of the following stages under `extends.parameters.stages`.
 
 <details><summary><b>2b. Publish stage using managed identity</b></summary>
 
+> Note: id tokens have a lifetime on the order of 1 hour. Because the esrp-npm-release script has no way to refresh this token,
+> if publishing packages takes more than 1 hour, this process may fail. Retrying the pipeline stage is safe and will start where
+> the previous attempt left off.
+
 ```yml
 - stage: publish
   displayName: Publish
--- a/packages/esrp-npm-release/README.md
+++ b/packages/esrp-npm-release/README.md
@@ -239,9 +239,9 @@ The typical release process using this tool has two main parts, which could be e
 
 In 1ES PT, the presence of a production release job applies stricter network isolation policies to the _entire_ pipeline. If you're using a single pipeline, you'll need to re-enable the access the rest of the pipeline needs (GitHub, Azure) via the `networkIsolationPolicy` setting.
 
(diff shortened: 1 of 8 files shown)


packages/esrp-npm-release/README.md
```diff
@@ -356,88 +356,195 @@ extends:
...
+
+            # Staging storage credentials
+            STAGING_STORAGE_ACCOUNT_NAME: <storage account name>
+            # set above by AzureCLI@2 but must be mapped in
+            STAGING_CLIENT_ID: $(STAGING_CLIENT_ID)
+            STAGING_TENANT_ID: $(STAGING_TENANT_ID)
+            STAGING_ID_TOKEN: $(STAGING_ID_TOKEN)
+
+            # ESRP credentials using managed identity
+            ESRP_ID_TOKEN: $(ESRP_ID_TOKEN)
```
Abe27342, line 539: A potential problem with using this approach is that unlike the certificate-based authentication approach, the esrp-npm-release script does not have a way to refresh associated tokens. Lifetime of this token is ~1 hour (as far as I understand, exact value can depend on tenant-specific entra policies). Experimentally, this approach is good enough for my team but it's admittedly cutting it a bit close (seems plausible if ESRP has high variance in publish times, publishing the full set of our largest release group could take above that though I haven't seen it yet).

An alternative method here would involve adding a dep on \@azure/msal-node or \@azure/identity (latter probably only if we have plans to move other bits of this such as kv secret fetching to \@azure libraries) and having the user inject information about the service connection itself + the pipeline's access token such that the running javascript could refresh the token as necessary.

I'm open to doing that either in place of this PR or as a follow-up, but it's inherently a bit of a larger change. Let me know what you think.

packages/esrp-npm-release/src/ESRPReleaseService.ts
```diff
@@ -105,7 +109,16 @@ export class ESRPReleaseService {
     this.#logger = params.logger;
     this.#clientId = params.clientId;
     this.#tenantId = params.tenantId;
-    this.#authCertificatePfx = params.authCertificatePfx;
+    if (params.authCertificatePfx && params.idToken) {
+      throw new ReleaseError('ESRP auth requires exactly one of authCertificatePfx or idToken');
+    }
+    if (params.idToken) {
+      this.#esrpAuth = { idToken: params.idToken };
```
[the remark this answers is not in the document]
Abe27342: discussion on this should go here: https://github.com/microsoft/beachball/pull/1362#discussion_r3707997357

packages/esrp-npm-release/README.md
ecraig12345: I think it's fine to leave it as-is for now (with a note about the potential issues) and only do a more complete fix if teams actually start hitting this issue. In theory, if a token does expire, it should be possible to re-run that stage of the pipeline, and with the ReleaseState logic it will skip the packages that were successful.

ecraig12345 (approved):
Thanks!

